PR #13, opened 2023-04-14
131

IngaSokman:


src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+import buttonArrowLightGrey from 'assets/icons/button_arrow_light_grey.svg'
+import buttonArrowDarkGrey from 'assets/icons/button_arrow_dark_grey.svg'
+import buttonArrowBlue from 'assets/icons/button_arrow_blue.svg'
+import classNames from 'classnames'
+import classes from './styles.module.scss'
+
+export interface ButtonProps extends BaseButtonProps {
+  appearance?: 'primary' | 'secondary' | 'text'
+  size?: 'm' | 's'
+  icon?: boolean
```

The button should be more abstract. Currently "icon" prop is a boolean and the button can only be rendered with the arrow icon. We should have a button that can be rendered with any icon. 

For example in our design we already have places, where the button is rendered with other icons. 

https://www.figma.com/file/VqCwED3Dymn7L1gGoAOK27/EKI-t%C3%B5lkev%C3%A4rav?node-id=427-46968&t=Q1I3r6YlCbuXJxB4-0

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+        return disabled ? buttonArrowLightGrey : buttonArrowBlue
+      }
+      default: {
+        return buttonArrowWhite
+      }
+    }
+  }, [appearance, disabled])
+
+  if (!icon) return null
+  return <img className={iconClass} src={iconButtonClass} alt="buttonArrow" />
```
KaarelKa: **[non-blocker]**

I would suggest using svg directly for button component. It should be passed in as a prop. 

If we use  tag and pass the prop in as src, we would need to pass multiple icons to every button (for example, secondary button would need both disabled icon and enabled icon).

If instead we use svg, then we can just change the color with css, which means that every button would only need 1 icon.

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export type IconProps = {
+  appearance?: 'primary' | 'secondary' | 'text'
+  icon?: boolean
+  disabled?: boolean
+}
+
+const Icon: FC<IconProps> = ({ icon, appearance, disabled }) => {
+  const iconClass = `${classes.icon} ${classes[`icon--${appearance}`]} `
```

1. In bürokrat project they are using global scss. For that they use BEM (https://getbem.com/) to name their classes. However we are using modular scss, so we wouldn't have to deal with the naming issues that global scss has, so lets create our own simple class names instead of following BEM.

2. Noticed that here you are constructing the class names string manually. We have a library called "classnames" in this project, which does it for us. Let's use classnames library for this (for future consistency, there is nothing wrong with manually constructing the string otherwise)

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+const Button: FC<PropsWithChildren<ButtonProps>> = ({
+  appearance = 'primary',
+  size = 'm',
+  icon,
+  hidden,
+  disabled,
+  children,
+  ...rest
+}) => {
+  const buttonClass = `${classes.btn} ${classes[`btn--${appearance}`]} ${
```
KaarelKa: Already commented above

src/components/molecules/Button/styles.module.scss
```diff
@@ -0,0 +1,117 @@
...
+}
+
+.buttonText {
+  order: 2;
+  @include buttonTextStyle;
+  line-height: $line-height-24;
+}
+
+.btn {
+  $self: &;
```
KaarelKa: I don't think we need $self if we stop using BEM. There will not be any need to construct long class names, rather just short ones for adding extra styles.

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from 'components/atoms/BaseButton/BaseButton'
+import buttonArrowWhite from 'assets/icons/button_arrow_white.svg'
+import buttonArrowLightGrey from 'assets/icons/button_arrow_light_grey.svg'
+import buttonArrowDarkGrey from 'assets/icons/button_arrow_dark_grey.svg'
+import buttonArrowBlue from 'assets/icons/button_arrow_blue.svg'
+import classNames from 'classnames'
+import classes from './styles.module.scss'
+
+export interface ButtonProps extends BaseButtonProps {
+  appearance?: 'primary' | 'secondary' | 'text'
```

It would be better to use Enums for for props that accept very specific strings. Those can also be imported separately, to avoid any typos. You can check the InputTypes enum for an example (both implementation and usage)

Idea is that we would import both Button and the Enum from here and when passing in either size or appearance prop we take it from the Enum.

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+        return buttonArrowWhite
+      }
+    }
+  }, [appearance, disabled])
+
+  if (!icon) return null
+  return <img className={iconClass} src={iconButtonClass} alt="buttonArrow" />
+}
+
+const Button: FC<PropsWithChildren<ButtonProps>> = ({
```
KaarelKa, line 46: **[blocker]**

It seems that the height of "secondary" button is 4px bigger than design

src/components/molecules/Button/styles.module.scss
```diff
@@ -0,0 +1,117 @@
...
+  appearance: none;
+  display: inline-flex;
+  align-items: center;
+  background: none;
+  border: 0;
```
KaarelKa: **[blocker]**

Some unnecessary styles here.

1. border: 0; -- it already is 0, so this is redundant
2. background: none; -- doesn't seem to do anything. Looks like this is code that should be in reset.scss, but we have an alternative of this there already (background-color: transparent). So this can be skipped as well

src/components/molecules/Button/styles.module.scss
```diff
@@ -0,0 +1,117 @@
...
+  text-decoration: none;
+  border-radius: $border-radius-20;
+  white-space: nowrap;
+  background-color: $color-primary;
+
+  &:focus {
+    outline: none;
+  }
+
+  &--disabled {
```
KaarelKa: BEM styling, was mentioned above. For disabled we actually don't need a new class name at all. Disabled is also passed to the wrapping "BaseButton" and can be targeted with the ":disabled" pseudo-selector.

src/components/molecules/Button/styles.module.scss
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  &--s {
+    padding: 5px 24px;
+    @include smallButtonTextStyle;
+  }
+}
+
+.icon {
```
KaarelKa: This will change, once we use svg directly anyway, but generally, when exporting icons from figma, then take the full 24px x 24px version. They all have a box around them and it will make it easier to deal with any icon. We will also not have any weird padding/width/height/margin values this way.

**[note]**

Extra note here is that we should use the "spacing" helper for all margin/padding/gap values to make sure alignment is correct everywhere.

src/components/molecules/Button/styles.module.scss
```diff
@@ -0,0 +1,117 @@
...
+@import 'src/styles/mixins';
+
+.baseButton {
+  display: flex;
+}
+
+.buttonText {
+  order: 2;
+  @include buttonTextStyle;
+  line-height: $line-height-24;
```
KaarelKa: **[blocker]**

Unnecessary new line-height. The correct line-height was already applied. Most text in this project has a line-height of 1.5, which is applied globally and is also true for buttonText.

buttonTextStyle has font-size of 16px. This means that line-height: 1.5 means we multiply 16 by 1.5, which is 24px.

KaarelKa (changes requested):

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
+import { FC, PropsWithChildren, SVGProps, FunctionComponent } from 'react'
```

FC and FunctionComponent are the same thing. FC is just a shorthand for it, so you won't need both. In the rest of the project FC is used so let's stick with FC for consistenct

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export type IconProps = {
+  icon?: FunctionComponent<SVGProps<SVGSVGElement>>
+  className?: string
+  ariaLabel?: string
+}
+
+const Icon: FC<IconProps> = ({ icon: IconComponent, className, ariaLabel }) => {
+  const ariaLabelToUse = ariaLabel
```

Let's avoid creating random variables just to copy an existing variable, if there is no need for it.

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  if (!IconComponent) return null
+  return (
+    <IconComponent
+      className={(classes.icon, className)}
```

1. classes.icon doesn't seem to exist
2. construction a class name like this won't work. Classname should be a string, not (string, string). Even if classes.icon would exist, it wouldn't be applied like this

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+  disabled,
+  className,
+  children,
+  ...rest
+}) => {
+  if (hidden) return null
+  return (
+    <BaseButton
+      className={classNames(
+        classes.btn,
```
KaarelKa, line 62: **[blocker]** Wondering why classes.btn and classes.baseButton are separated here ? They don't seem to be used anywhere else.

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+        classes[`${appearance}`],
```

Both "appearance" and "size" props are already strings, there is no need to use template literals here to parse them into string.

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+        classes[`${size}`],
+        classes.baseButton,
+        className
+      )}
+      disabled={disabled}
+      {...rest}
+    >
+      <span className={classes.buttonText}>{children}</span>
+      <Icon icon={icon} className={className} ariaLabel={ariaLabel} />
```

Not sure if there is a case, where we need to pass the same className to both the outer container (BaseButton) and the icon inside it. However even if there is, it should be 2 separate props. Currently it is nearly impossible to use the className prop in any meaningful way. For example the classic usage for it would be to position the component. In this case we would be positioning both the button component and the icon inside the button at the same time, which would break the button.

The intuitive behaviour for the "className" prop is that it is always passed to the outer most container of the component. That way it will be used the same way as the className prop of native html elements.

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <BaseButton
+      className={classNames(
+        classes.btn,
+        classes[`${appearance}`],
+        classes[`${size}`],
+        classes.baseButton,
+        className
+      )}
+      disabled={disabled}
```

Let's use spreading for props, which are not used directly, but just passed on to the child component. It will be easier to find actual uses/implementation of a prop in the project this way.

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+      {...rest}
+    >
+      <span className={classes.buttonText}>{children}</span>
```
KaarelKa, line 70: **[blocker]**

There seems to be no spacing between the icon and text. Doesn't match design

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,77 @@
...
+  icon?: FunctionComponent<SVGProps<SVGSVGElement>>
+  className?: string
+  ariaLabel?: string
+}
+
+const Icon: FC<IconProps> = ({ icon: IconComponent, className, ariaLabel }) => {
+  const ariaLabelToUse = ariaLabel
+
+  if (!IconComponent) return null
+  return (
```

Some issues with the icon:

1. Color of the icon seems to be wrong, at least in secondary button.
2. Seems that the icon is in front of the text for secondary button and after the text in primary. We should have the capability to have the icon on either side of the text for any button, but this also means that by default we should always choose 1 side (either before or after the text).

src/components/molecules/Button/styles.module.scss
```diff
@@ -0,0 +1,104 @@
+@import 'src/styles/variables';
```
KaarelKa: **[blocker]**

I think you don't need to have the "src/" part when importing styles. It should work with just "styles/variables"

src/components/molecules/Button/styles.module.scss
```diff
@@ -0,0 +1,104 @@
...
+
+.baseButton {
+  display: flex;
+}
+
+.buttonText {
+  order: 2;
+}
+
+.m {
```
KaarelKa: **[blocker]**

Currently button sizes don't seem to be correct. At least the small button height is off. You are defining .m and .s styles here, but overwriting them with the .secondary or .primary classes, so any overlapping properties will just use whatever the classes for "appearance" dictate.

src/styles/variables.scss
```diff
@@ -31,52 +31,56 @@ $font-size-12: 12px;
 // Line heights
 $line-height: 1.5;
 $line-height-14: 14px;
+$line-height-24: 24px;
```
KaarelKa: **[blocker]**

Unused variable

KaarelKa (changes requested):

src/components/molecules/Button/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+  ariaLabel?: string
+}
+
+const Icon: FC<IconProps> = ({ icon: IconComponent, className, ariaLabel }) => {
+  if (!IconComponent) return null
+  return <IconComponent className={className} aria-label={ariaLabel} />
+}
+
+const Button: FC<PropsWithChildren<ButtonProps>> = ({
+  appearance = 'primary',
```
KaarelKa, line 47: **[blocker]**

Primary button focused style does not match design.

KaarelKa (changes requested):

src/components/molecules/Button/Button.tsx
IngaSokman: Yes, missed that - fixed

KaarelKa (approved):

PR #14, opened 2023-04-26
130 date picker

IngaSokman:


src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <div
+          className={classNames(
+            classes.wrapper,
+            message && classes.errorMessage
+          )}
+        >
+          <DatePicker
+            id="DatePicker"
+            selected={value ? new Date(formattedValue) : null}
+            dateFormat={dateFormat}
```

Currently the default value for dateFormat seems to be 'MM.dd.yyyy', however we will be using 'dd.MM.yyyy' nearly everywhere (possibly even everywhere), so it would be more convenient to make 'dd.MM.yyyy' the default value here, to avoid passing 1 extra prop every time we want to use DatePicker

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { ReactComponent as Calender } from 'assets/icons/calender.svg'
+import classNames from 'classnames'
+
+import 'react-datepicker/dist/react-datepicker.css'
+import classes from './styles.module.scss'
+
+export interface DatePickerInputProps {
+  name: string
+  className?: string
+  error?: FieldError
```
KaarelKa, line 21: **[blocke]**

looks like "error" prop is not used here, instead you are using "message" prop for the same purpose. This means that for this input we need to handle all error manually, while others get the error from react-hook-form.

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+      placeholder,
+      dateFormat,
+      className,
+      ariaLabel,
+      message,
+      value,
+      ...rest
+    } = props
+
+    const { onChange } = rest
```
KaarelKa: **[non-blocker]** why is this taken out separately ?

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+        >
+          {label}
+        </Label>
+        <div
+          className={classNames(
+            classes.wrapper,
+            message && classes.errorMessage
+          )}
+        >
+          <DatePicker
```
KaarelKa, line 74: **[blocker]**

Currently when I pick any date in the calendar (for example 11.05.2023), the corresponding day (11.) will be marked on every month in the calendar

Screenshot 2023-05-04 at 08 00 53

src/components/molecules/DatePickerInput/styles.module.scss
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    &:focus {
+      outline: none;
+      color: $color-text-dark;
+      border-color: $color-primary;
+    }
+  }
+}
+
```
KaarelKa, line 50: **[non-blocker]**

Not 100% sure, but I think you could have the :disabled selector in the same place, where you have :focus and ::placeholder selectors

KaarelKa (changes requested):

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -0,0 +1,103 @@
...
+        </Label>
+        <div
+          className={classNames(classes.wrapper, error && classes.errorMessage)}
+        >
+          <DatePicker
+            id="DatePicker"
+            selected={value ? new Date(formattedValue) : null}
+            dateFormat={'dd.MM.yyyy'}
+            locale="et-EE"
+            filterDate={isWeekday}
```

Just noticed that weekend dates are disabled by default. This doesn't match with our usecases according to design. While having the opportunity to disable some dates is good, they should not be disabled by default (unless the project only has datepickers, where those days are disabled).

KaarelKa (changes requested):

src/components/molecules/DatePickerInput/DatePickerInput.tsx
IngaSokman: Yes, good point overlooked it. Removed the default option and changed style accordingly.

PR #15, opened 2023-05-03
change displaying keycloak user, fix fetching authentication on refresh

thenouan:


.env.example
```diff
@@ -1,3 +1,3 @@
-REACT_APP_KEYCLOAK_realm = Tolkevarav DEV
```
KaarelKa, line 1: Noticed that you removed these from .env.example. Are these no longer needed ? If they are, then it would be good to keep the values for testing here for local FE development

KaarelKa (changes requested):

.env.example
thenouan: The reason why I removed those is that it gives out information to others, since repos are public. e.g. someone sees this url and decides to make a ddos attack etc.

PR #16, opened 2023-05-09
109

KaarelKa:


IngaSokman:
Is there a design for mobile views? Noticed that the padding and overflow is off on mobile. Otherwise approved

IngaSokman (approved):

PR #17, opened 2023-05-09
130 time picker

IngaSokman:


KaarelKa:
Seems to have some merge conflicts now, should be synced with dev branch

src/components/atoms/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,79 @@
+import { useEffect, useState } from 'react'
```
KaarelKa: **[blocker]** this shouldn't be an atom. Basically the logic is that atoms can't consist of other atoms. If we have a component that consists of atoms, then that should already be a molecule and this one uses 2 atoms already

src/components/atoms/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+import BaseButton from 'components/atoms/BaseButton/BaseButton'
+import { ReactComponent as ButtonArrow } from 'assets/icons/button_arrow.svg'
+import Icon from '../Icon/Icon'
```
KaarelKa: **[blocker]**

Let's stick with absolute imports for cases like this.

Relative imports for components should be used, when importing some smaller part of this component and only if that part is only used inside this component + is situated in the same component folder (basically if that part will always be moved with this component).

src/components/atoms/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }, [selectedTime, setValue])
+
+  const controlBottom = () => {
+    if (selectedTime !== end - 1) {
+      return setSelectedTime(selectedTime + 1)
+    }
+    if (selectedTime === end - 1) {
+      return setSelectedTime(start)
+    }
+    return setSelectedTime(start)
```
KaarelKa: **[blocker]**

If I'm not mistaken, then code will never reach this default return.  "selectedTime !== end - 1" covers all cases, other than "selectedTime === end - 1" so there will be no condition left, which would pas by both "if" statements

src/components/atoms/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+      return setSelectedTime(end - 1)
+    }
+  }
+
+  return (
+    <div className={classes.control}>
+      <div className={classes.timeContainer}>
+        <div className={classes.selector} />
+        <BaseButton onClick={controlTop} className={classes.controlTimeTop}>
+          <Icon icon={ButtonArrow} ariaLabel={'top time arrow'} />
```
KaarelKa: **[blocker]**

Every text we use should be translatable. If we keep these hardcoded strings inside the codebase, it will be a pain to find them later on. Only time these are allowed, is when the component/feature is WIP.

src/components/atoms/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+    if (selectedTime === end - 1) {
+      return setSelectedTime(start)
+    }
+    return setSelectedTime(start)
+  }
+
+  const controlTop = () => {
+    if (selectedTime !== start) {
+      return setSelectedTime(selectedTime - 1)
+    } else {
```
KaarelKa: **[blocker]**

No need for else after return, however here returning sth doesn't really make sense, so rather keep the else and remove the returns

src/components/atoms/TimeColumn/TimeColumn.tsx
KaarelKa: Also returning the call of a function here is not really necessary. We just need to call out the function, not return whatever the call of that function returns

src/components/atoms/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <div className={classes.selector} />
+        <BaseButton onClick={controlTop} className={classes.controlTimeTop}>
+          <Icon icon={ButtonArrow} ariaLabel={'top time arrow'} />
+        </BaseButton>
+        <div className={classes.timeWrapper}>
+          {map(timeArray, (time) => {
+            return (
+              <div
+                key={time}
+                className={`${
```
KaarelKa: **[blocker]** We have added classNames library for adding conditional classes. Let's use it, instead of constructing the string manually, to keep the codebase consistent

src/components/atoms/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+    if (selectedTime !== end - 1) {
+      return setSelectedTime(selectedTime + 1)
+    }
+    if (selectedTime === end - 1) {
+      return setSelectedTime(start)
+    }
+    return setSelectedTime(start)
+  }
+
+  const controlTop = () => {
```
KaarelKa: **[blocker]**

Possibly - Disccuss this with Martine to make sure, but to me it seems that up arrow should increase the number and down arrow should decrease the number, not the other way around.

src/components/atoms/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+    if (selectedTime !== start) {
+      return setSelectedTime(selectedTime - 1)
+    } else {
+      return setSelectedTime(end - 1)
+    }
+  }
+
+  return (
+    <div className={classes.control}>
+      <div className={classes.timeContainer}>
```
KaarelKa: **[blocker]** Do we need 2 containing divs here with just 1 child ? In most cases this is not needed, unless you are doing sth very specific. In this case I think we can easily skip 1 of them

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -43,51 +107,53 @@ const DatePickerInput = forwardRef<HTMLInputElement, DatePickerInputProps>(
...
+          >
+            {label}
+          </Label>
+          <div
+            className={classNames(
+              classes.wrapper,
+              error && classes.errorMessage
+            )}
+          >
+            <DatePickerComponent
```

Let's split Datepicker and Timepicker into 2 separate inputs with their own types. They have very little in common and this will be quite hard to maintain. 

However you can create a new helper component out of the shared part (content container, label, wrapper and errormessage). Both Timepicker and Datepicker can then use that helper component as a wrapper and we will avoid code duplication without ending up with a messy component

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -43,51 +107,53 @@ const DatePickerInput = forwardRef<HTMLInputElement, DatePickerInputProps>(
...
-          />
-          <Icon
-            icon={Calender}
-            className={classNames(disabled && classes.disabledCalender)}
-            ariaLabel={ariaLabel}
-          />
-          <InputError {...error} />
+        <div className={classes.contentContainer}>
+          <Label
+            htmlFor="DatePicker"
```

Currnently clicking on the label of any time or date input on the page will result in the first date picker of the page being focused. Every label should be attached the the corresponding input and should only focus that input.

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -5,12 +5,13 @@ import DatePicker, {
   registerLocale,
```
KaarelKa, line 4: **[blocker]** new console error added with timepicker, always check for new console errors/warning before creating a PR and fix them.


Screenshot 2023-05-15 at 10 43 19

src/components/molecules/DatePickerInput/DatePickerInput.tsx
KaarelKa: Will review these again after the split, currently not checking any further

src/components/molecules/DatePickerInput/styles.module.scss
```diff
@@ -6,67 +6,73 @@
...
 
+  .contentContainer {
+    display: flex;
+  }
+
   .label {
     @include labelTextStyle;
     display: inline-block;
     padding-right: spacing(1.25);
+    padding-top: 6px;
```
KaarelKa: **[blocker]** Generally padding and margin values should use "spacing", however in this case you are using padding to center the label. We should instead use layout css for this. Currently this is a flex container with direction row, so align-items should be used

KaarelKa: We do this, because it will be much easier to make the components responsive once the time comes. It will mean less changes, if we for example have different font-size/line-height/input height on smaller screens or they are just changed some time in the future

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -43,51 +107,53 @@ const DatePickerInput = forwardRef<HTMLInputElement, DatePickerInputProps>(
...
+              value={value}
+              disabled={disabled}
+              ariaLabel={ariaLabel}
+              ariaLabelToUse={ariaLabelToUse}
+              placeholder={placeholder}
+              timePicker={timePicker}
+              onChange={onChange}
+              {...rest}
+            />
+            <TimePickerInput
```
KaarelKa: **[blocker]** seems that accessibility support is not great for these either. Every component should support interaction with just the keyboard as well. Currently timepicker can't be opened with just the keyboard

src/components/molecules/TimePickerInput/TimePickerInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+  disabled?: boolean
+  ariaLabel?: string
+  value?: string
+  timeColumnVisibility: () => void
+  inputRef?: Ref<HTMLInputElement> | null
+  error?: FieldError
+  showSeconds?: boolean
+}
+
+const TimeInput: FC<TimeInputProps> = forwardRef(
```
KaarelKa: **[blocker]** forwardRef not used correctly here. You are wrapping TimeInput with forwardRef, however the forwarded ref is not used. Instead the ref is passed as a prop (inputRef) and passed from there. Let's stick with using forwardRef, not ref as a prop.

src/components/molecules/TimePickerInput/TimePickerInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+const TimePickerInput = ({
+  onChange,
+  disabled,
+  ariaLabel,
+  value,
+  error,
+  showSeconds,
+  timePicker,
+}: TimePickerInputProps) => {
+  const [hour, setHour] = useState<string>('00')
```
KaarelKa: **[blocker]** I think we are using too many internal states currently. hour, minute and second value are currently kept in 3 places at once (Inside TimePickerInput, inside the form and inside TimeColumn).

Ideally we want to keep the value only in 1 place to avoid issues that come from data duplication. In this case it is kind of okay to have it in 2 places, but 3 seems overkill

src/components/molecules/TimePickerInput/TimePickerInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+  const [minute, setMinute] = useState<string>('00')
+  const [second, setSecond] = useState<string>('00')
+  const [isTimeColumnOpen, setTimeColumnOpen] = useState<boolean>(false)
+
+  const timeColumnVisibility = () => {
+    setTimeColumnOpen((prevState) => !prevState)
+  }
+
+  const handleClickOutside = (event: MouseEvent) => {
```
KaarelKa: **[blocker]**

We will most likely use outside click logic in many places in this project, so would be good to abstract this away or use some 3rd party hook for it, so we can reuse the logic easily. 1 option would be to use https://ahooks.js.org/hooks/use-click-away 

Extra note is that currently you are only handling mouse clicks, which means that this won't work for people who are using a keyboard instead of a mouse. Support for this can be added quite easily with https://ahooks.js.org/hooks/use-click-away or here, by just adding a few extra event listeners

src/components/molecules/TimePickerInput/TimePickerInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+  error,
+  showSeconds,
+  timePicker,
+}: TimePickerInputProps) => {
+  const [hour, setHour] = useState<string>('00')
+  const [minute, setMinute] = useState<string>('00')
+  const [second, setSecond] = useState<string>('00')
+  const [isTimeColumnOpen, setTimeColumnOpen] = useState<boolean>(false)
+
+  const timeColumnVisibility = () => {
```
KaarelKa: **[blocker]**

Naming wise it would be good to use verbs for function names. It will make the codebase more predictable and easier to read (every verb does something, which means they are some type of method/function)

src/components/molecules/TimePickerInput/TimePickerInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+      <div
+        className={
+          !isTimeColumnOpen || disabled
+            ? classes.hiddenContainer
+            : classes.container
+        }
+        ref={timeColumnRef}
+      >
+        <TimeColumn start={0} end={24} value={hour} setValue={setHour} />
+        <TimeColumn start={0} end={59} value={minute} setValue={setMinute} />
```
KaarelKa: **[blocker]** Can't pick 59 minutes or 59 seconds right now

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -24,13 +25,76 @@ export interface DatePickerInputProps {
   placeholder?: string
   disabled?: boolean
   value?: string
+  timePicker?: boolean
```
KaarelKa: **[blocker]** a lot of duplicated types here. Everything that is defined under DatePickerComponentProps is also defined under DatePickerInputProps, so DatePickerInputProps could just extend DatePickerComponentProps.

Only exception is ariaLabelToUse, but this is not exactly necessary and is also not correctly defined. It is not required by DatePickerComponent, however it is always passed, since ariaLabel is required by DatePickerInput. There is also no need to pass both ariaLabel and ariaLabelToUse here, you will not need both, so just ariaLabel will be enough.

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -24,13 +25,76 @@ export interface DatePickerInputProps {
...
+        {...rest}
+        onChange={handleDateChange}
+      />
+      <Icon
+        icon={Calender}
+        className={classNames(
+          classes.dateIcon,
+          disabled && classes.disabledCalender
+        )}
+        ariaLabel={ariaLabel}
```
KaarelKa: **[blocker]** icons don't need aria labels. aria-label is meant for interactive elements: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

src/components/molecules/TimePickerInput/TimePickerInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+} from 'react'
+import TimeColumn from 'components/atoms/TimeColumn/TimeColumn'
+import { ReactComponent as Clock } from 'assets/icons/clock.svg'
+import { FieldError } from 'react-hook-form'
+import Icon from 'components/atoms/Icon/Icon'
+import classNames from 'classnames'
+
+import classes from './styles.module.scss'
+
+export type TimePickerInputProps = {
```
KaarelKa: **[blocker]**

Let's remove some type repetition from here as well. It can be quite confusing, since it's hard to tell what props have the same type between parent and child (which ones are passed directly from parent to child) and which ones are new/modified

KaarelKa (changes requested):

src/components/atoms/TimeColumn/TimeColumn.tsx
IngaSokman: Yes, true, this default return is redundant, removed it

src/components/atoms/TimeColumn/TimeColumn.tsx
IngaSokman: Good point, asked Martine just to be sure, waiting for reply currently

IngaSokman: It should be the other way around - top arrow buttons increase the number and down arrows decrease the number, changed it

src/components/atoms/Icon/Icon.tsx
```diff
@@ -3,12 +3,11 @@ import { FC, SVGProps } from 'react'
 export type IconProps = {
```
KaarelKa: **[blocker]**

Hmm I'm wondering what is the point of this wrapper component. This seems to be meant for icons as svgs, however those already have the className prop. Se essentially this component doesn't really seem to do anything.
It currently just add 1 extra import + some extra complexity, but no real value.

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -4,34 +4,82 @@ import DatePicker, {
...
 }
 
+export type DatePickerInputProps = SharedDateProps & {
+  name: string
+  className?: string
+  error?: FieldError
+  label?: JSX.Element | string
+}
+
+export type DatePickerComponentProps = SharedDateProps
```
KaarelKa: **[non-blocker]**

Seems a bit unnecessary to create a new variable here, it is essentially const a = b

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -4,34 +4,82 @@ import DatePicker, {
...
+type SharedDateProps = {
   ariaLabel: string
   placeholder?: string
   disabled?: boolean
   value?: string
   onChange: (value: string) => void
 }
 
+export type DatePickerInputProps = SharedDateProps & {
+  name: string
```
KaarelKa, line 22: **[blocker]**

These seem to be duplicated types. In reality DatePickerInputProps seem to be a combination of SharedProps (or rather DatePickerComponentProps) and InputWrapperProps. You could import InputWrapperProps here instead of duplicating. This would also make it a lot clearer for other developers that the parent component (DatePickerInput) accepts props for InputWrapper and DatePickerComponent.

Only prop inside InputWrapperProps, which is not here is children, however that can just be omitted

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -47,49 +95,23 @@ const DatePickerInput = forwardRef<HTMLInputElement, DatePickerInputProps>(
...
-      return onChange(changeDateToString(value))
-    }
-
-    const splittedValue = value?.split('/')
-    const formattedValue =
-      splittedValue?.[2] + '-' + splittedValue?.[1] + '-' + splittedValue?.[0]
-
     return (
-      <Field
+      <InputWrapper
```
KaarelKa, line 94: **[blocker]**

Previously we had the issue where clicking on the label always opened the first DatePicker, but now clicking on the label doesn't open any datepicker.

Basically for all input fields, clicking on the label should focus on the input. So for datepicker it should open the datepicker.

This is built into html as well. If you have any input component with some id, then you can bind any label component to it. https://www.w3schools.com/tags/att_label_for.asp

Basically you can bind them like this:

<label htmlFor="randomInput">text</label>
<input id="randomInput" />

For other inputs (Like checkbox or TextInput, this works because, because @radix-ui/react-form does this automatically. their <Label> and <Control> elements are bound, however we are not using Control here, so this should be done manually

src/components/molecules/TimeColumn/TimeColumn.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  return (
+    <div className={classes.control}>
+      <div className={classes.selector} />
+      <BaseButton onClick={controlTop} className={classes.controlTimeTop}>
+        <Icon icon={ButtonArrow} />
+      </BaseButton>
+      <div className={classes.timeWrapper}>
+        {map(timeArray, (time) => {
```
KaarelKa: **[blocker]**

Would prefer a different implementation here.

1. Effectively we are just changing the number, so there is no need to create so many divs. Currently it is creating 60 separate divs for minutes, when we could just have 1 div tag, which displays value formatted correctly.

2. All text should be in appropriate tags. divs should never contain text directly, we have a list of htmls tags meant specifically for different types of text in different contexts

3. Can ask from Martine about this, but when interacting with this input I find myself wanting to use by keyboard to write the numbers in, instead of pressing the arrow button 30 times (if I want to select 16:30 time for example), so I would make these into input fields. We should still have the arrows, but the option to edit the number manually would make it much more user friendly.

src/components/molecules/TimePickerInput/TimePickerInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+        minute
+      )}:${formatTimeString(second)}`
+      const formattedTime = `${formatTimeString(hour)}:${formatTimeString(
+        minute
+      )}`
+      onChange(showSeconds ? timeWithSeconds : formattedTime)
+    }, [hour, minute, second, onChange, showSeconds])
+
+    return (
+      <InputWrapper
```
KaarelKa, line 156: **[blocker]**

Same issue as with DatePicker, can't focus, when pressing on label

KaarelKa (changes requested):

KaarelKa (approved):

KaarelKa:
@IngaSokman Approved, but has merge conflict. You can merge this after fixing the conflicts

PR #19, opened 2023-05-15
132

KaarelKa:
**Note** Currently left placeholders for language picker and user dropdown select, because we need dropdown select component for those https://github.com/keeleinstituut/tv-tolkevarav/issues/261

**Extra note** second level menu items look very weird on collapsed menu. This is currently waiting for an answer from designer in figma

Also the spacing increment was changed to 4px to better match our design. I went over all current uses of spacing and changed the values there


IngaSokman:
**[blocker]** From the side navigation the "?" icon and "Kasutusjuhend" text are missing

src/components/organisms/SideBar/SideBar.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }, [navCollapsed, setNavCollapsed])
+  return (
+    <nav
+      className={classNames(
+        classes.sidebarContainer,
+        navCollapsed && classes.collapsed
+      )}
+    >
+      <BaseButton
+        className={classNames(classes.listItem)}
```
IngaSokman: **[non-blocker]** Redundant classNames

IngaSokman:
**[blocker]** Do I understand correctly that the "Sätted" menu item should have always top and bottom border even when the submenu items are not visible? Also, when the "Sätted" button is clicked the top and bottom border are doubled - I think it's because of the outline style:
Screenshot 2023-05-15 at 15 08 32

src/components/organisms/Header/styles.module.scss
```diff
@@ -0,0 +1,35 @@
+@import 'styles/variables';
+
+.header {
+  width: 100%;
+  background: $color-white;
+  height: $header-height;
+  box-shadow: 0px 0px 2px rgba(0, 0, 0, 0.14), 0px 2px 2px rgba(0, 0, 0, 0.12), 0px 1px 3px rgba(0, 0, 0, 0.2);
```
IngaSokman: **[non-blocker]** You could remove the "px" units when the value is zero

KaarelKa:
@IngaSokman Yeah, the settings menu should always have the borders and doubling is currently the only thing I could read out of the styleguide. There is no specific design for focused style for this component, so I applied the same focused style as other menu items have.

This :focus style is mainly meant for accessibility reasons (Has to be clear, which item has focus, so it's easier to move around the page using keyboard)

Actually I'll switch this to :focus-visible selector, since it is meant for keyboard users, not people who use the mouse

IngaSokman (approved):

PR #21, opened 2023-05-15
259

IngaSokman:


src/components/atoms/Modal/Modal.tsx
```diff
@@ -0,0 +1,58 @@
+import { FC, ReactNode, PropsWithChildren } from 'react'
+import * as Dialog from '@radix-ui/react-dialog'
+import classNames from 'classnames'
+import Button, { AppearanceTypes } from 'components/molecules/Button/Button'
```
KaarelKa: **[blocker]**

Please go over Atomic Design again, there seems to be some confusion about this in the last few PR's. 
Very briefly:

1. Atoms - basic building blocks
2. Molecules - Collections of 1 or more atoms. Usually more complex and actively used components. (For example Input component, which uses Label, input, errormessage atoms, etc)
3. Organism - Collection of 1 or more molecules and 0 or more atoms. Usually big complex components, like a Form, which essentially includes all different types of inputs + some internal logic

The following we use more freely:

4. Templates - Basically sections of pages. Contain some organisms/molecules/atoms and include the internal logic for that section, when relevant, so we can reuse them for different pages. We also use these for wrappers and layouts.
5. Pages - In a sense we have pages as well based on atomic design, but here we mostly just focus on pages of a website, so these can consist of basically anything relevant for that page.

src/components/atoms/Modal/Modal.tsx
```diff
@@ -0,0 +1,58 @@
...
+  trigger,
+  children,
+  customDialogContent,
+  topButton,
+}) => {
+  const { t } = useTranslation()
+
+  return (
+    <Dialog.Root>
+      <div>{trigger}</div>
```
KaarelKa: **[blocker]** (Possibly, maybe not)

Since trigger and footer are React Nodes, then do we need to wrap them in <div> ? Or is this needed for something else ?

src/components/atoms/Modal/Modal.tsx
```diff
@@ -0,0 +1,58 @@
+import { FC, ReactNode, PropsWithChildren } from 'react'
```
KaarelKa: **[blocker]**

Currnetly it seems that we only have 1 type of modal semi-implemented, which is the narrow modal, with wrong bottom spacing + with the wrong close button (The close button that should be used on another type of modal).

This components should be much easier to use, currently whenever we want to use it we need to pass in the full footer component, the full header component + a bunch of extra styles to create the modal we want. This then has to be done every time we want to use a modal. 

I would prefer if more of that work would be done inside this component. 

We essentially have:

1. 3 different sizes for modal (narrow, medium and big), currently the component only supports the narrow version out of the box. Would be good if this could just be picked with 1 prop, instead of passing in a bunch of styles, a new footer and new header.

2. 3 different types of footers (buttons split, buttons on the right, no footer). Currently we also need to manually create the component every time and pass it in, which will get quite troublesome, since we have a lot of Modals in this project. Would be good if this could also be switched with a prop. Possibly with extra props for button labels + what they will do.

3. Close button either visible or not (this seems to be handled)


1 option would be to rewrite this component to match and another option would be to basically use this as a modal basis and creating a new Modal component ,which we will actually use, which itself uses the modal bases and maps out all the the different types of modals we will have.

KaarelKa (changes requested):

src/components/atoms/Modal/Modal.tsx
IngaSokman: Yes, we don't need to wrap them, removed

src/components/atoms/Modal/Modal.tsx
IngaSokman: Totally overlooked, automatically added it to atoms..

src/components/atoms/Modal/Modal.tsx
IngaSokman: Yes, made the modal more easier to use. Ready for review

src/components/molecules/Button/Button.tsx
```diff
@@ -63,6 +63,7 @@ const Button: FC<PropsWithChildren<ButtonProps>> = ({
         classes[appearance],
         classes[size],
         classes[iconPositioning],
+        icon ? classes.customButtonPadding : '',
```
KaarelKa: **[non-blocker]**

One reason why we use the "classnames" library is because it will remove all falsey values, so there is no need to return empty string here, any falsey value is fine. This means that you could just write it as 
icon && classes.customButtonPadding

src/components/molecules/Button/styles.module.scss
```diff
@@ -100,14 +100,23 @@
   @include buttonTextStyle;
   padding: spacing(2) spacing(10);
 
+  &.customButtonPadding {
```
KaarelKa: **[blocker]**

Play through all button variants with and without icon with these changes. Currently sth seems to be wrong. For example the primary button we have as an example has wrong height, when it has an icon. But go over other buttons as well with all appearance + size + with/without icon combos to make sure we don't have any other similar cases

src/components/organisms/Modal/Modal.tsx
```diff
@@ -0,0 +1,132 @@
...
+  breakButtonLabel?: string
+  proceedButtonLabel?: string
+  handleClose: () => void
+  handleProceed: () => void
+  open: boolean
+  setOpen: Dispatch<SetStateAction<boolean>>
+  titleFont?: TitleFontTypes
+}
+
+export interface ModalFooterProps {
```
KaarelKa, line 55: **[blocker]** 

All of these props are passed directly from the parent component, so you could avoid duplication of type definitions here by having ModalProps interface extend ModalFooterProps.
Extra benefit is that it will be much more clear to developers that these are exactly the same props

src/components/organisms/Modal/Modal.tsx
```diff
@@ -0,0 +1,132 @@
...
+const ModalFooter: FC<ModalFooterProps> = ({
+  buttonsPosition = 'right',
+  breakButtonLabel,
+  proceedButtonLabel,
+  handleClose,
+  handleProceed,
+}) => {
+  return (
+    <div className={classes[buttonsPosition]}>
+      <Button appearance={AppearanceTypes.Secondary} onClick={handleClose}>
```
KaarelKa: **[blocker]**

It seems that currently we only support modals with exactly 0 buttons or 2 buttons, where the left button is of Secondary type and the right button is primary.

According to design we have some cases that are not covered by this. 
For example: https://www.figma.com/file/VqCwED3Dymn7L1gGoAOK27/EKI-t%C3%B5lkev%C3%A4rav?type=design&node-id=2346-107207&t=je6Y5mYgtspPolA1-4 

here we have split buttons, where the button the left is a text button and the button on the right is secondary button.

**Extra note**:

Currently we also only support passing onClick action to these buttons however in the same example the left button is actually a link, so we want to pass "href" to it instead of onClick

Another example which is not supported is:

https://www.figma.com/file/VqCwED3Dymn7L1gGoAOK27/EKI-t%C3%B5lkev%C3%A4rav?type=design&node-id=936-91678&t=Z4cUh8GLdliLnTqX-4

Here 1 of the buttons will be disabled until some times are picked, which means that it needs some information from the content of the modal to decide whether it is disabled or not.



**Potential solution**

Based on this I would suggest keeping the wrapper for the buttons for their positioning inside this component, however allowing the passing of the Buttons as props. Basically just 1 prop, which is rendered inside this wrapper div. It can have as many buttons as we want (possibly event 1 or 3 for the future, or whatever number). This wrapper will just decide how they are positioned + how much spacing they have from edges and from top/bottom + makes sure that they will always be visible at the bottom of the modal

src/components/organisms/Modal/Modal.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  return (
+    <Dialog.Root open={open} onOpenChange={setOpen}>
+      {trigger}
+      <Dialog.Portal>
+        <Dialog.Overlay className={classes.dialogOverlay} />
+        <Dialog.Content
+          className={classNames(classes.dialogContent, classes[size])}
+        >
+          {topButton && (
```
KaarelKa: **[blocker]**

inline conditional rendering for something this short is normally fine, however in this case we are using it for a component (<Button>), which has the "hidden" prop, so there is no need to use inline conditional rendering here. Always prefer using the "hidden" prop, if it exists

src/components/organisms/Modal/Modal.tsx
```diff
@@ -0,0 +1,132 @@
...
+            <Button
+              icon={Close}
+              appearance={AppearanceTypes.Secondary}
+              onClick={handleClose}
+              className={classes.topButton}
+            >
+              {t('button.cancel')}
+            </Button>
+          )}
+          <Dialog.Title
```
KaarelKa: **[blocker]**

Currently we only support having a string title with 2 different font styles and it is always at the top of the modal, however we have cases, where instead of the title we have some other component. Like the progress bar here: 

https://www.figma.com/file/VqCwED3Dymn7L1gGoAOK27/EKI-t%C3%B5lkev%C3%A4rav?type=design&node-id=834-80371&t=Z4cUh8GLdliLnTqX-4

src/components/organisms/Modal/Modal.tsx
```diff
@@ -0,0 +1,132 @@
...
+  customDialogContent?: string
+  topButton?: boolean
+  size?: ModalSizeTypes
+  buttonsPosition: ButtonPositionTypes
+  breakButtonLabel?: string
+  proceedButtonLabel?: string
+  handleClose: () => void
+  handleProceed: () => void
+  open: boolean
+  setOpen: Dispatch<SetStateAction<boolean>>
```
KaarelKa, line 50: **[non-blocker]** (for now at least)

I would suggest leaving this type a bit more open. Currently our modal can only be used, if we have some react setState function to pass to setOpen, however we are currently not 100% if we will have this every time. We might use some other state management for holding the state of modal in some rare cases.

KaarelKa (changes requested):

src/components/organisms/Modal/Modal.tsx
IngaSokman: Ok, got it, I'll change it then when we have more information

src/components/molecules/Button/Button.tsx
```diff
@@ -52,6 +52,7 @@ const Button: FC<PropsWithChildren<ButtonProps>> = ({
   hidden,
   className,
   children,
+  disabled,
```

1. disabled prop was already passed to BaseButton inside ...rest, so taking it out here and passing it manually will make it slightly harder to search for actual implementation of disabled for button + it adds some extra code here.

2. I also noticed now that ButtonProps interface defines the "disabled" prop again, even though it already extends BaseButtonProps, which have the type for "disabled" defined.

src/components/molecules/Button/Button.tsx
```diff
@@ -63,8 +64,11 @@ const Button: FC<PropsWithChildren<ButtonProps>> = ({
         classes[appearance],
         classes[size],
         classes[iconPositioning],
+        icon && classes.customButtonPadding,
+        appearance === 'text' && icon && classes.textIconSize,
```
KaarelKa: **[non-blocker]**

Not a big issue, however this className could be skipped. We already have all the information for this. Basically if both classes.text and classes.customButtonPadding are added to button, then it means that appearance === 'text' && icon, so you could just move this part to scss (&.text { &.customButtonPadding } ).

Extra not here is that "customButtonPadding" class name should be changed. This is actually very specific padding which we add, when there is an icon. "customButtonPadding" is very generic and doesn't really contain any information, other than that we are dealing with some class name that changes padding

src/components/organisms/Modal/Modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+  None = 'none',
+}
+
+export enum TitleFontTypes {
+  Gray = 'gray',
+  Black = 'black',
+}
+
+export interface ModalButtonProps extends ButtonProps {
+  onClick: () => void
```
KaarelKa: **[blocker]** 

I don't think onClick should be required. For example if we have a modal, which should open a link in a new tab, then there will be no need to pass any onClick function

src/components/organisms/Modal/Modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+const ModalFooter: FC<PropsWithChildren<ModalFooterProps>> = ({
+  buttonsPosition = 'right',
+  buttons,
+}) => {
+  return (
+    <div className={classes[buttonsPosition]}>
+      {map(buttons, (button, index) => {
+        return (
+          <Button
+            key={index}
```
KaarelKa: **[blocker]**

Would be better to use prop spreading here. It seems like we have a lot of code, which is not necessary + will create some confusion. Whenever a prop is taken out and manually passed on, then the developer will need to check why this is done. Usually the reason is that it is somehow modified, or used somewhere else. In this case most of the props are not used for anything else, so this will create some future overhead for developers, when maintaining the project.

Extra overhead in cases like this would be 

1. If a new prop is added to Button or BaseButton, then we also need to go over the entire project and check every place where props are passed to Button or BaseButton manually and pass the new prop there as well. With spreading these props we can just define it in BaseButton and automatically everything that extends our BaseButton accepts this prop.

2. When we want to see what is done with some prop, then usually what we do is we use find over the entire project and check everything it is passed through. This means that all these files, where the props are passed manually will pop up, which will clutter the search with irrelevant results

3. Unnecessary code = more code = harder to maintain

src/components/organisms/Modal/Modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+            disabled={button?.disabled}
+            size={button?.size}
+            iconPositioning={button?.iconPositioning}
+            icon={button?.icon}
+            ariaLabel={button?.ariaLabel}
+            hidden={button?.hidden}
+            onClick={button?.onClick}
+            href={button?.href}
+            className={classNames(
+              buttonsPosition !== 'spaceBetween' && classes.modalButton,
```
KaarelKa: **[non-blocker]**

Not too important here as well, but basically this classname could also be skipped. you can target the button directly from classes.right. ( .right { button { ...css } } ).

src/components/organisms/Modal/Modal.tsx
KaarelKa: This is currently not resolved yet, since the "title" prop can only be a string and ower progressbar will not be a string.

src/components/organisms/Modal/styles.module.scss
```diff
@@ -0,0 +1,92 @@
...
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 100%;
+  max-height: 100%;
+  padding: spacing(10) 0;
+}
+
+.dialogTitle {
+  padding: 0 spacing(17.5) spacing(4) spacing(17.5);
```
KaarelKa: **[non-blocker]**

padding shorthand also work with 3 values, so the last spacing(17.5) here is redundant

KaarelKa (changes requested):

KaarelKa (approved):

KaarelKa:
@IngaSokman some merge conflicts, but after fixing those you can merge this PR

PR #24, opened 2023-05-22
133 table features

VaabelMaret:


src/components/molecules/Button/Button.tsx
```diff
@@ -65,6 +66,7 @@ const Button: FC<PropsWithChildren<ButtonProps>> = ({
         icon && classes.customIconPadding,
         className
       )}
+      disabled={disabled}
```

Was already passed in ...rest, no need to take it out and pass it manually, since it's not used for anything else inside this file

package.json
```diff
@@ -3,9 +3,11 @@
   "version": "0.1.0",
```
KaarelKa, line 3: **[blocker]**

Getting a ton of warnings in terminal

Screenshot 2023-05-24 at 17 56 01

Also I see a lot of console.log-s are still left, these should all be removed before creating a PR

src/components/atoms/Container/Container.tsx
```diff
@@ -0,0 +1,20 @@
+import { FC, PropsWithChildren } from 'react'
+import classNames from 'classnames'
+import classes from './styles.module.scss'
+
+export type ContainerProps = {
```
KaarelKa: **[non-blocker]**

1. No need to export a type, if it is not used anywhere else.
2. For simple types it would be good to prefer "interface" for consistency, however this is not too important right now, will do some cleanup after our current deadline and then we will start enforcing it

src/components/organisms/TableFilter/TableFilter.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { ReactComponent as FilterIcon } from 'assets/icons/filter.svg'
+import Button, {
+  AppearanceTypes,
+  SizeTypes,
+  IconPositioningTypes,
+} from 'components/molecules/Button/Button'
+
+export type FilterProps = {
+  column: Column<any, unknown>
```
KaarelKa: **[blocker]**

Project should generally have 0 uses of any (unless it's for a very good reason).
I can see that Column already accepts type arguments, so we should pass actual type arguments here, based on what type of data we have

(also exported type for something that is not used anywhere else)

src/components/organisms/TableFilter/TableFilter.tsx
```diff
@@ -0,0 +1,54 @@
...
+  table: Table<any>
+}
+
+const TableFilter: FC<FilterProps> = ({ column, table }) => {
+  const { t } = useTranslation()
+
+  const firstValue = table
```
KaarelKa: **[blocker]**

A couple of console warning from here, unused variables. If these are just leftover code that won't be used, then you can remove them. If these are something that we need in the future, then a good approach would be to comment them out + add some TODO: comment with information about what these will be used for

src/components/organisms/TableFilter/TableFilter.tsx
```diff
@@ -0,0 +1,54 @@
...
+      </datalist>
+
+      <Button
+        // onClick={(value) => column.setFilterValue(value)}
+        appearance={AppearanceTypes.Text}
+        size={SizeTypes.S}
+        icon={FilterIcon}
+        ariaLabel={t('label.filter')}
+        iconPositioning={IconPositioningTypes.Right}
+        placeholder={`Search... (${column.getFacetedUniqueValues().size})`}
```
KaarelKa: **[blocker]**

If this placeholder will remain and this isn't some WIP code, then we should use translatable strings for everything right away. It will be a lot harder to go through the entire project later to find all the strings we need to replace

If this is not sure yet, or this will be removed, then leave a TODO: comment here with information about what needs to be done/what it depends on.

(by TODO: comment i mean a comment that starts with "TODO:", our project is already full of them, so it will be easy to go through the entire project and check all the TODO items later"

src/components/organisms/TablePagination/TablePagination.tsx
```diff
@@ -0,0 +1,110 @@
...
+  return (
+    <div className={classes.paginationWrapper}>
+      <div className={classes.pagination}>
+        <Button
+          appearance={AppearanceTypes.Text}
+          size={SizeTypes.S}
+          icon={Arrow}
+          ariaLabel={t('label.button_arrow')}
+          iconPositioning={IconPositioningTypes.Left}
+          onClick={() => previousPage()}
```
KaarelKa: **[blocker]**

prefer passing the function directly to onClick, instead of creating a new anonymous function during every rerender
 onClick={previousPage} instead of  onClick={() => previousPage()}

src/components/organisms/TablePagination/TablePagination.tsx
```diff
@@ -0,0 +1,110 @@
...
+        >
+          <ul className={classes.links}>
+            {[...Array(getPageCount())].map((_, index) => (
+              <li
+                key={`${id}-${index}`}
+                className={classNames({
+                  [classes.active]: getState().pagination.pageIndex === index,
+                })}
+              >
+                <Link
```
KaarelKa: **[blocker]**

Anything clickable should use our BaseButton component. Ideally Button, but if that's not possible, then at least BaseButton.

src/components/organisms/TablePagination/TablePagination.tsx
```diff
@@ -0,0 +1,110 @@
...
+          icon={Arrow}
+          ariaLabel={t('label.button_arrow')}
+          iconPositioning={IconPositioningTypes.Left}
+          onClick={() => previousPage()}
+          disabled={!getCanPreviousPage()}
+        />
+
+        <nav
+          role="navigation"
+          aria-label={t('label.pagination_navigation') || ''}
```
KaarelKa: **[blocker]**

Let's remove the || '' from this aria-label, then we will get typescript errors, if the translation for "label.pagination_navigation" is missing for some reason.

src/components/organisms/TablePagination/TablePagination.tsx
```diff
@@ -0,0 +1,110 @@
...
+            ))}
+          </ul>
+        </nav>
+        <Button
+          appearance={AppearanceTypes.Text}
+          size={SizeTypes.S}
+          icon={Arrow}
+          ariaLabel={t('label.button_arrow')}
+          iconPositioning={IconPositioningTypes.Left}
+          onClick={() => {
```
KaarelKa: **[blocker]**

Same issue here, no need to create a new anonymous function

src/components/organisms/TablePagination/TablePagination.tsx
```diff
@@ -0,0 +1,110 @@
...
+          }}
+          disabled={!getCanNextPage()}
+          className={classes.arrow}
+        />
+      </div>
+      <div className={classes.pageSizeWrapper}>
+        <label htmlFor={id} className={classes.pageSizeLabel}>
+          {t('label.pagination_result_count')}
+        </label>
+        <select
```
KaarelKa: **[non-blocker]**

Would be good to have a comment here, that this should be replaced with our select component

src/components/organisms/TablePagination/styles.module.scss
```diff
@@ -0,0 +1,92 @@
+@import 'styles/variables';
+@import 'styles/mixins';
+
+// Pagination
+.paginationWrapper {
+  display: flex;
+  justify-content: center;
+  padding: 8px 44px;
```
KaarelKa: **[blocker]**

all padding/margin/gap values and all other positioning values should use the "spacing" helper

src/components/organisms/TablePagination/styles.module.scss
```diff
@@ -0,0 +1,92 @@
...
+    gap: 15px;
+    margin: 0 auto;
+
+    + .pageSizeWrapper {
+      margin-left: 0;
+    }
+
+    button {
+      height: 100%;
+      &:focus {
```
KaarelKa: **[blocker]**

Applies for the entire table. Currently it seems that all or nearly all elements we can interact with in the table have no focus style. This is an accessibility issue. Every components should be fully usable with just the keyboard and focus styles are necessary, so the user understand what they are interacting with.

src/components/organisms/TablePagination/styles.module.scss
```diff
@@ -0,0 +1,92 @@
...
+    .arrow {
+      transform: rotate(-180deg);
+    }
+  }
+  .pageNumber,
+  .pageSizeLabel,
+  .pageSizeSelect {
+    @include subtitleSmallTextStyle;
+    color: $color-input-main;
+    line-height: $line-height-21;
```
KaarelKa: **[blocker]**

unnecessary line-height here. We already have global line-height at 1.5.
subtitleSmallTextStyle has font-size of 14px, which means it's line-height is 14 * 1.5 = 21px already

src/components/organisms/TablePagination/styles.module.scss
```diff
@@ -0,0 +1,92 @@
...
+    grid-template-columns: max-content 47px;
+    align-items: center;
+    justify-items: center;
+    gap: 8px;
+    margin-left: auto;
+
+    .pageSizeSelect {
+      width: 100%;
+      appearance: none;
+      padding: spacing(0.5) spacing(4) spacing(0.5) spacing(1.5);
```
KaarelKa: **[non-blocker]**

paddings seem to be off from design here, however we will replace pageSizeSelect with our select component anyway, so not too important for now

src/components/templates/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+import TablePagination from 'components/organisms/TablePagination/TablePagination'
+
+export enum TableSizeTypes {
+  L = 'l',
+  M = 'm',
+  S = 's',
+}
+
+type DataTableProps = {
+  data: any
```
KaarelKa: **[blocker]**

Getting rid of all the "any" types should start from here. our DataTable component should accepts at least 1 type argument and most of the other types can be inferred from that. 

You can find infromation about this under "Generics" in typescript documentation.

You can also check the example we have with our DynamicForm and DynamicInputComponent about how to use the type argument and how to pass it

src/components/templates/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+  tableSize: string
+}
+
+type ColumnMeta = {
+  meta: {
+    size: number | string
+  }
+}
+
+declare module '@tanstack/table-core' {
```
KaarelKa: **[non-blocker]** (maybe)

Do we need to overwrite these types in @tanstack/table-core ? It seems to work without these overwrites as well, however not sure, did a very brief check for this

src/components/templates/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+  sortable,
+  filterable,
+  pagination,
+  setPagination,
+  globalFilter,
+  setGlobalFilter,
+  meta,
+}) => {
+  const [sorting, setSorting] = useState<SortingState>([])
+  const [columnFilters, setColumnFilters] = React.useState<ColumnFiltersState>(
```
KaarelKa: **[blocker]**

if useState is already imported from React, then there is no need to use React.useState. And let's stick with importing useState from react in this project for consistency.

src/components/templates/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+  title,
+  sortable,
+  filterable,
+  pagination,
+  setPagination,
+  globalFilter,
+  setGlobalFilter,
+  meta,
+}) => {
+  const [sorting, setSorting] = useState<SortingState>([])
```
KaarelKa: **[blocker]**

it seems that currently this table has some internal sorting logic added, however our sorting will mostly rely on BE requests, because FE won't know about information on other pages, so it can't sort reliably. We might have some separate logic for sorting, if all the data is on 1 page/FE already has all the info, but if we do, then that should also come from outside of the table component. 

So here clicking on the sorting button should call some function that is passed into DataTable as a prop and shouldn't care about what that function does.

(Same is true for filtering)

src/components/templates/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+    getExpandedRowModel: getExpandedRowModel(),
+  })
+
+  return (
+    <Container>
+      <h4 className={classes.title}>{title}</h4>
+      <div className={classes.tableWrapper}>
+        <table className={classNames(classes.dataTable, classes[tableSize])}>
+          <thead>
+            {table.getHeaderGroups().map((headerGroup) => (
```
KaarelKa: **[non-blocker]**

Would suggest some code splitting here to make this more readable.
Like table.getHeaderGroups().map((headerGroup) => <HeaderGroup {...props} />) or sth like this

src/components/templates/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+                {headerGroup.headers.map((header) => {
+                  return (
+                    <th
+                      key={header.id}
+                      style={{
+                        width: (header.column.columnDef as CustomColumnDef).meta
+                          ?.size,
+                      }}
+                    >
+                      {header.isPlaceholder ? null : (
```
KaarelKa: **[blocker]**

This is definitely too long for inline conditional rendering + this also has nested inline conditional rendering.
Basically same rules apply as in RA - nested inline conditional rendering is always a no and other inline conditional rendering is fine, if it's short enough and doesn't hurt readability

src/components/templates/DataTable/TestingTable.tsx
```diff
@@ -0,0 +1,356 @@
+import { FC, useState, HTMLProps, useEffect, useRef } from 'react'
```
KaarelKa: **[blocker]**

Let's change the folder structure a bit.


1. Let's create a "tables" folder inside "organisms" or "templates" which will contain all the specific tables (like TestingTable and UsersTable).
2. Each table should consist of 1. folder named the same as component (Component), inside that folder a .tsx file with the same name (Component.tsx) + possibly a styles.module.scss file, if needed.
3. The general "DataTable" component should not be in the general "tables" folder, it can be inside "organisms" separately

src/components/templates/DataTable/TestingTable.tsx
```diff
@@ -0,0 +1,356 @@
...
+  progress: number
+  subRows?: Person[]
+}
+
+function IndeterminateCheckbox({
+  indeterminate,
+  className = '',
+  ...rest
+}: { indeterminate?: boolean } & HTMLProps<HTMLInputElement>) {
+  const ref = useRef<HTMLInputElement>(null!)
```
KaarelKa: **[blocker]**

typescript error here, you can fix this, by removing the "null" from useRef constructor and using useImperativeHandle for populating the values of ref.current with some custom values. (example usage inside organisms/modals/index)

src/components/templates/DataTable/TestingTable.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+  useEffect(() => {
+    if (typeof indeterminate === 'boolean') {
+      ref.current.indeterminate = !rest.checked && indeterminate
+    }
+  }, [ref, indeterminate])
+
+  return (
```
KaarelKa: **[blocker]**

This should use our CheckBox component, current version doesn't seem to match design (the checked version) + doesn't have the focused highlight.

src/components/templates/DataTable/TestingTable.tsx
```diff
@@ -0,0 +1,356 @@
...
+    <input
+      type="checkbox"
+      ref={ref}
+      className={className + ' cursor-pointer'}
```
KaarelKa: **[blocker]**

1. Couldn't find this class name being used anywhere, what is it used for ? 
2. Let's use the "classnames" library to construct the classes string for consistency

src/components/templates/DataTable/TestingTable.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+const columnHelper = createColumnHelper<Person>()
+
+const columns = [
+  columnHelper.accessor('id', {
+    header: () => 'Id',
+    cell: ({ row, getValue }) => (
+      //Example for link row cell
+      <div
+        style={{
```
KaarelKa: **[blocker]** (potentially) do we need inline styles for this div or can we use className ?

src/components/templates/DataTable/styles.module.scss
```diff
@@ -0,0 +1,80 @@
...
+    width: 100%;
+    text-align: left;
+    margin-bottom: 0;
+    display: table;
+    @include bodyTextStyle;
+  }
+
+  .l th,
+  td {
+    padding: 12px 14px;
```
KaarelKa: **[blocker]**

mentioned above as well, but a lot of values in this file that should use spacing

src/components/templates/DataTable/styles.module.scss
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  button {
+    width: max-content;
+    padding: 0;
+    svg g path,
+    svg path {
+      fill: $color-input-main;
+    }
+    &:focus {
```
KaarelKa: **[blocker]**

focus style mentioned above

src/styles/variables.scss
```diff
@@ -31,6 +31,8 @@ $font-size-12: 12px;
 // Line heights
 $line-height: 1.5;
 $line-height-14: 14px;
+$line-height-24: 24px;
```
KaarelKa: **[blocker]**

Both of these are unnecessary. They are used in places, where the line-heigh is already correct (the 1.5 line-height we have globally).

If there is something in the table that is overwriting our global line-height ,then you can use the $line-height variable here, instead of $line-height-24 and line-height-21

KaarelKa (changes requested):

IngaSokman:
There's padding missing between checkbox and "First name". 
Screenshot 2023-06-19 at 12 25 30

IngaSokman:
The sorting icon is smaller than in the design. Also, I see that the icon is one svg element, just for clarification - should it be one svg element or two separate icons? In Bürokratt I see that the usage is also different when the icon is being clicked.
Screenshot 2023-06-19 at 12 35 02

IngaSokman:
In Veera and in EKI UI I see that the filter icon's color is Primary/Black Coral, #5D6071, should't it be not blue?
Screenshot 2023-06-19 at 12 49 24

IngaSokman:
In the design different arrow icon is being used:
Screenshot 2023-06-19 at 14 00 24

src/components/templates/Tables/ExamplesTable/styles.module.scss
```diff
@@ -0,0 +1,7 @@
+.row {
+  display: grid;
+  grid-auto-flow: column;
+  width: max-content;
+  gap: 10px;
+  align-items: left;
```
IngaSokman: Invalid property value

src/components/organisms/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+export enum TableSizeTypes {
+  L = 'l',
+  M = 'm',
+  S = 's',
+}
+
+type DataTableProps<TData extends RowData> = {
+  data: TData[]
+  columns: ColumnDef<TData>[]
+  tableSize: string
```
IngaSokman: **[blocker]** Let's use TableSizeTypes when defining tableSize, this way we ensure that only valid values from the TableSizeTypes enum can be assigned to it.

src/components/organisms/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,118 @@
...
+      rowSelection,
+      expanded,
+      ...{ pagination },
+    },
+    meta,
+    ...(pagination && { getPaginationRowModel: getPaginationRowModel() }),
+    onPaginationChange: setPagination,
+    getCoreRowModel: getCoreRowModel(),
+    enableRowSelection: true, //enable row selection for all rows
+    // enableRowSelection: row => row.original.age > 18, // or enable row selection conditionally per row
```
IngaSokman: If both enable for all rows and enable for per row are neede, maybe it would be best to make it a prop.

src/components/organisms/TableHeaderGroup/TableHeaderGroup.tsx
```diff
@@ -0,0 +1,127 @@
...
+          icon={SortingArrows}
+          ariaLabel={t('label.button_arrow')}
+          iconPositioning={IconPositioningTypes.Left}
+          className={classes.sortingButton}
+        />
+      )}
+
+      {flexRender(column.columnDef.header, header.getContext())}
+
+      {!isEmpty(filterOption) && onColumnFiltersChange ? (
```
IngaSokman: **[non-blocker]** Would use logical and operator here as well.

src/components/organisms/DataTable/styles.module.scss
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  .l th,
+  td {
+    padding: spacing(3) spacing(3.5);
+  }
+  .m th,
+  td {
+    padding: spacing(2) spacing(3.5);
+    line-height: 20px;
```
IngaSokman: **[blocker]** Add the line height as a variable. Also, is this needed?

src/components/organisms/TablePagination/styles.module.scss
```diff
@@ -0,0 +1,76 @@
...
+
+        .pageNumber {
+          padding: spacing(0);
+          display: flex;
+          align-items: center;
+          justify-content: center;
+          width: 25px;
+          height: 25px;
+          border-radius: 50%;
+          background-color: white;
```
IngaSokman: **[blocker]** Add all colors from variables

src/components/organisms/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+type DataTableProps<TData extends RowData> = {
+  data: TData[]
+  columns: ColumnDef<TData>[]
+  tableSize: TableSizeTypes
+  title?: string
+  onSortingChange?: (value: string | boolean, columnId: string) => void
+  onColumnFiltersChange?: (filters: string[], columnId: string) => void
+  pagination?: PaginationState
+  setPagination?: Dispatch<SetStateAction<PaginationState>>
```
KaarelKa: **[blocker]**

Would prefer if we didn't use the Dispatch<SetStateAction... types for functions. This limits the prop that we can pass here heavily. Basically this type tells us that this can only be used with react useState, which is an unnecessary limitation and in case of table there is a pretty big chance that we won't be storing the pagination data inside useState.

In similar cases prefer regular arrow function types. (state: PaginationState) => void for example

src/components/organisms/TableHeaderGroup/TableHeaderGroup.tsx
```diff
@@ -0,0 +1,128 @@
...
+      setCurrentIndex(newIndex)
+    }
+  }, [id, newIndex, onSortingChange, value])
+
+  if (hidden) return null
+
+  return (
+    <div className={classes.headingWrapper}>
+      {!isEmpty(sortingOption) && (
+        <Button
```
KaarelKa: **[blocker]**

While these smaller inline conditional rendering statements are not much of an issue, they should still not be used, if the underlying component has the "hidden" prop defined. Our Button has the "hidden" prop, so would be better to use that instead.

src/components/organisms/TableHeaderGroup/TableHeaderGroup.tsx
```diff
@@ -0,0 +1,128 @@
...
+          onClick={handleOnSorting}
+          appearance={AppearanceTypes.Text}
+          size={SizeTypes.S}
+          icon={value === 'asc' ? SortingLess : SortingMore}
+          ariaLabel={t('label.button_arrow')}
```
KaarelKa: **[blocker]**

ariaLabel should rather contain information about what the button does, not what it looks like. Exact content will be added by client/translator, however currently it seems you are using the same translation key 'label.button_arrow' for a lot of different ariaLabels, so the correct values can't be added currently.

**Note*** Also check other ariaLabels you have added

src/components/organisms/TableHeaderGroup/TableHeaderGroup.tsx
KaarelKa: Or not use inline conditional rendering at all, since hidden prop is defined in the TableColumnFilter component and should be used instead

src/components/organisms/TableHeaderGroup/styles.module.scss
```diff
@@ -0,0 +1,29 @@
...
+    width: max-content;
+    svg g path,
+    svg path {
+      fill: $color-input-main;
+    }
+    svg {
+      padding: 0;
+    }
+    &:focus {
+      box-shadow: inset 0 0 0 2px transparent;
```
KaarelKa: **[blocker]**

Discussed this in the office, but you should contact Martine and discuss this part again. Would prefer not to break Accessbility for something that might look a bit better visually. Every component that the user can interact with should have some visible focused style

src/components/organisms/TablePagination/TablePagination.tsx
```diff
@@ -0,0 +1,112 @@
...
+          onClick={previousPage}
+          disabled={!getCanPreviousPage()}
+          className={classes.arrows}
+        />
+
+        <nav role="navigation" aria-label={t('label.pagination_navigation')}>
+          <ul className={classes.links}>
+            {[...Array(getPageCount())].map((_, index) => (
+              <li
+                key={`${id}-${index}`}
```
KaarelKa: **[blocker]**

Go over React docs for useId hook. https://react.dev/reference/react/useId This should not be used to generate keys for a list. Guessing this is because the id will change every time TablePagination is rerendered, which means that if you also use it as a key for some list of elements inside TablePagination, all of them will also be rerendered, since their key has changed.

src/components/organisms/TablePagination/TablePagination.tsx
```diff
@@ -0,0 +1,112 @@
...
+        name={id}
+        ariaLabel={t('label.pagination_result_count')}
+        label={t('label.pagination_result_count')}
+        options={pageSizeOptions}
+        value={toString(getState().pagination.pageSize)}
+        onChange={(value) => {
+          setPageSize(Number(value))
+        }}
+        placeholder={toString(getState().pagination.pageSize)}
+        multiple={false}
```
KaarelKa: **[non-blocker]**

redundant prop, multiple should be false by default.

src/components/organisms/TableColumnFilter/TableColumnFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+        className={classes.iconButton}
+      />
+
+      <DropdownContent
+        name={name}
+        ariaLabel={name}
+        options={filterOption}
+        multiple
+        buttons
+        cancelButtonLabel={t('button.cancel')}
```
KaarelKa: You can go over this again, after merging dev into your branch. Removed some props from DropdownContent

src/components/organisms/TablePagination/styles.module.scss
```diff
@@ -0,0 +1,79 @@
...
+        }
+      }
+    }
+
+    .arrows {
+      padding: spacing(2);
+      &.toRight {
+        transform: rotate(-180deg);
+      }
+      &:focus {
```
KaarelKa: **[blocker]** mentioned above as well about focus styles

src/components/templates/Tables/AddedUsersTable/AddedUsersTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [pagination, setPagination] = useState<PaginationState>({
+    pageIndex: 0,
+    pageSize: 10,
+  })
+  const tableData = users?.data
+
+  const usersData = useMemo(() => {
+    return (
+      map(tableData, (data) => {
+        const arrayOfRoles = map(data.roles, ({ name }) => name)
```
KaarelKa: **[non-blocker]**

With lodash map, if you just want 1 key from the array you can use the shorthand: map(data.roles, 'name')

KaarelKa (changes requested):

KaarelKa:
@VaabelMaret Make sure you merge the latest dev into your branch. Just merged some stuff to dev, which affects your code a bit as well

KaarelKa (approved):

PR #29, opened 2023-05-24
261

IngaSokman:


src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,121 @@
+import { FC } from 'react'
```
KaarelKa: **[blocker]**

Same issue we had with date and time picker. Clicking on the label doesn't open the selection

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,121 @@
...
+    } else {
+      onChange(selectedOption ? selectedOption?.value : '')
+    }
+  }
+
+  return (
+    <>
+      <div
+        className={classNames(classes.dropdownMenu, classes[dropdownSize])}
+        hidden={disabled || !isOpen || !!error}
```
KaarelKa: **[blocker]**

This means that if the user gets some error from here, for example forgets to fill the field and it's required, they won't have any way of fixing that error, since we won't allow the opening of the dropdown anymore

src/components/organisms/DropdownContent/DropdownContent.tsx
KaarelKa: **[blocker]**

Some mismatches from design:

1. Tags are not aligned with the input component
2. Label alignment is off, when we have some tags

Design: 

Screenshot 2023-05-25 at 09 51 56

Reality: 

Screenshot 2023-05-25 at 09 53 41

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+const DropdownContent: FC<DropdownContentProps> = ({
+  dropdownSize = 'l',
+  disabled,
+  isOpen,
+  error,
+  searchInput,
+  options,
+  multiple = false,
```
KaarelKa, line 25: **[blocker]**

1. For multi select the dropdown closes right after I make 1 change, making it very inconvenient to make multiple selections.

2. Currently it seems that the selected values from multiselect are saved to the form right away, which renders the "Filtreeri" button useless. Should confirm with Martine how that button is suppose to behave.

3. "Tühista" button doesn't seem to do anything other than closing the dropdown. Not sure if this is correct behaviour, should be confirmed with Martine.

When looking at the component, I would assume that all the selections I make are temporary and when I press the submit button, they will be stored, when I press the cancel button they will be reset to previous state. Although not sure how closing the dropdown by some other method should work in this case.

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,121 @@
...
+        className={classNames(classes.dropdownMenu, classes[dropdownSize])}
+        hidden={disabled || !isOpen || !!error}
+      >
+        <div hidden={!searchInput}>{searchInput}</div>
+
+        {map(options, (option, index) => {
+          const isSelected = value && includes(value, option?.value)
+
+          return (
+            <li
```
KaarelKa: **[blocker]**

<li> elements should only be used inside an <ul>, <ol> or <menu> element https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,121 @@
...
+              key={index}
+              className={classes.dropdownMenuItem}
+              onClick={() => handleOptionSelect(option)}
+            >
+              {multiple && (
+                <CheckBoxInput
```
KaarelKa: **[blocker]**

Currently this breaks the logic we have with checkBoxes, where clicking on the label also ticks/unticks the checkbox. In that case handleOptionSelect is called twice. We should use the build in "onChange" of CheckBoxInput here, instead of wrapping it inside a component which manually tries to do the same thing

src/components/organisms/SelectionControlsInput/SelectionControlsInput.tsx
```diff
@@ -0,0 +1,156 @@
...
+  label?: JSX.Element | string
+  ariaLabel: string
+  value?: string
+  options: {
+    label: string
+    value: string
+  }[]
+  onChange: (value: string | string[]) => void
+  disabled?: boolean
+  defaultLabel?: string
```
KaarelKa: **[blocker]**

Let's change the name of this prop to "placeholder" so it would match our other input components

src/components/organisms/SelectionControlsInput/SelectionControlsInput.tsx
```diff
@@ -0,0 +1,156 @@
...
+    multiple = false,
+    helperText,
+    buttons = false,
+    cancelButtonLabel,
+    proceedButtonLabel,
+    searchInput,
+    dropdownSize,
+    tags = false,
+  },
+  ref
```
KaarelKa, line 65: **[blocker]**

ref doesn't seem to be used, so this forwardRef is redundant. Although before removing it I would suggest going over react-hook-form docs, so you understand what it uses ref for (It's not necessary, but then we have to handle some of the same things manually)

src/components/organisms/SelectionControlsInput/SelectionControlsInput.tsx
```diff
@@ -0,0 +1,156 @@
...
+    includes(value, option?.value)
+  )
+
+  const selectedOptionLabels = map(selectedOptionObjects, ({ label }) => label)
+
+  const singleSelectMenuLabel = value ? selectedOptionLabels : defaultLabel
+
+  const dropdownMenuLabel = multiple ? defaultLabel : singleSelectMenuLabel
+
+  return (
```
KaarelKa, line 89: **[non-blocker]**

It would take a bit of changes, but I think we could easily use the InputWrapper component here as well. It is around 95% the same code and other differences could probably be handled by just passing a className to InputWrapper.

See if you can make it work. If not, then let me know where the issue is and maybe we can figure it out

KaarelKa (changes requested):

src/components/molecules/InputWrapper/InputWrapper.tsx
```diff
@@ -12,21 +12,45 @@ export type InputWrapperProps = {
...
-  function InputWrapper({ label, name, error, className, children }, ref) {
+  function InputWrapper(
+    {
+      label,
+      name,
+      error,
+      className,
+      children,
+      onClick,
+      wrapperSizeClass,
```
KaarelKa: [non-blocker]

I would maybe just rename this to "wrapperClass" because we can add any styles to the wrapper with it, not just size specific styles.

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,150 @@
...
+              onClick={handleOnSave}
+              className={classes.dropdownButton}
+            >
+              {proceedButtonLabel}
+            </Button>
+          </div>
+        </ul>
+      </div>
+
+      <p hidden={!helperText} className={classes.helperText}>
```
KaarelKa, line 134: **[blocker]**

Noticed that this helpertext ruins the positioning of error. Error should be under the input field itself

 
Screenshot 2023-06-01 at 08 34 05

KaarelKa: Extra note -- I did some minor improvements to the error message + the testing form we have in Components page, would suggest pulling those changes in as well

src/components/organisms/SelectionControlsInput/SelectionControlsInput.tsx
KaarelKa: **[blocker]**

ref still seems to be in the wrong place. Currently when selection input is the first input in the form and I get an error on it after submitting, it won't be focused. Extra note here is that I wasn't exactly right, when I said that each html element can be focused. It's still basically true, but it's not true by default. Would suggest reading a bit more about how focus works:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus


In case of selection we want to focus on an element that would allow for the opening of the dropdown with the press of 1 button on keyboard afterwards.

src/components/organisms/SelectionControlsInput/SelectionControlsInput.tsx
```diff
@@ -0,0 +1,146 @@
...
+        </p>
+        <DropdownArrow
+          className={classNames(
+            disabled && classes.disabledDropdownIcon,
+            isOpen && !error && classes.openDropdownIcon
+          )}
+        />
+      </BaseButton>
+
+      <DropdownContent
```
KaarelKa, line 120: **[blocker]**

Go over the list of props here, I see that some stuff passed to DropdownContent here are not actually used inside DropdownContent

src/components/organisms/SelectionControlsInput/styles.module.scss
```diff
@@ -0,0 +1,76 @@
...
+  height: 40px;
+  text-align: left;
+  overflow: hidden;
+
+  &:focus {
+    border: 1px solid $color-primary;
+  }
+
+  &.m svg g path {
+    fill: #09090b;
```
KaarelKa: **[blocker]**

All colors should come from variables

KaarelKa (changes requested):

src/components/atoms/InputError/InputError.tsx
```diff
@@ -9,14 +9,22 @@ interface InputErrorProps {
...
   className?: string
+  selectionsError?: string
 }
 
-const InputError: FC<InputErrorProps> = ({ type, message, className }) => {
+const InputError: FC<InputErrorProps> = ({
+  type,
+  message,
+  className,
+  selectionsError,
```
KaarelKa: **[blocker]**

Do we need selectionsError class here ? It seems that we already have the "className" Prop here, which does exactly the same thing

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,151 @@
...
+      ? [...selectedValue]
+      : []
+    const optionIndex = selectedValues.indexOf(selectedOption)
+
+    const newSelectedValues =
+      optionIndex === -1
+        ? [...selectedValues, selectedOption]
+        : selectedValues.filter((value) => value !== selectedOption)
+
+    setSelectedValue(newSelectedValues)
```
KaarelKa, line 57: **[non-blocker]** 

Not important, but would suggest using lodash here to reduce the code inside this function.
Essentially here you are checking if the string (selectedOption) is inside the array values (selectedValues).

1. Array.isArray(selectedValue) - This check should not be necessary, typescript already says that this has to be an array of strings, so there shouldn't be any possibility that this isn't an array.

2. Currently you are using: const optionIndex = selectedValues.indexOf(selectedOption) + optionIndex === -1 to determine, if the value is already represented in the array or not. This is fine, but you could also use "includes" from lodash instead, to make this slightly shorter

3. Currently you are using filter for the array to remove the value. This is also fine, but would be slightly shorter with lodash without method.

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,151 @@
...
+  cancelButtonLabel,
+  proceedButtonLabel,
+  onChange,
+  setIsOpen,
+  helperText,
+  selectedOptionObjects,
+  tags,
+}) => {
+  const [selectedValue, setSelectedValue] = useState<string[]>(
+    value ? (Array.isArray(value) ? value : [value]) : []
```
KaarelKa: **[blocker]**

With typescript you have defined that value is either a string or undefined, so it should never be an array. This means that 

1. Currently this check is redundant, since we know what the type is 
2. Or this is typed incorrectly

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,151 @@
...
+    if (setIsOpen) {
+      setIsOpen(true)
+    }
+  }
+
+  const handleOnSave = () => {
+    onChange(selectedValue)
+  }
+
+  const handleCancel = () => {
```
KaarelKa, line 68: **[blocker]**

Currently this does not seem to work correctly. You are clearing the internal state of DropdownContent only.

1. You are changing the defaultvalues passed to an empty array, which means "Tühista" doesn't actually reset the changes the user made, but rather acts as "clear all".

2. Since you are only calling this on the internal state of DropdownContent, then the value in the form will still remain correct. Currently this means that we are displaying different information inside the DropdownContent and under the input, where we show selected values

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -0,0 +1,151 @@
...
+            const isMultiSelected =
+              selectedValue && includes(selectedValue, option?.value)
+            const isSingleSelected = value && includes(value, option?.value)
+
+            return (
+              <li key={index} className={classes.dropdownMenuItem}>
+                {multiple && (
+                  <CheckBoxInput
+                    name={name}
+                    ariaLabel={ariaLabel}
```
KaarelKa: **[blocker]**

Not ideal, this would mean that every option has the same ariaLabel, which can confuse users who rely on this a lot. Essentially for them every option would be the same. Rather use the option.label value for this

src/components/organisms/SelectionControlsInput/styles.module.scss
```diff
@@ -0,0 +1,76 @@
...
+    border: 1px solid $color-primary;
+    border-radius: 5px;
+  }
+
+  &.m svg g path {
+    fill: $color-text-dark;
+  }
+
+  &.s {
+    border: 1px solid #d2d3d8;
```
KaarelKa: **[blocker]**

Colors should come from variables

KaarelKa (changes requested):

PR #32, opened 2023-06-01
264

IngaSokman:


package.json
```diff
@@ -13,19 +13,21 @@
     "@types/node": "18.14.2",
```
KaarelKa, line 13: **[blocker]**

Currently the popup that opens is pushing content, instead of opening ontop of content.

src/components/organisms/DragAndDrop/DragAndDrop.tsx
```diff
@@ -0,0 +1,172 @@
...
+} from 'components/molecules/Button/Button'
+import { map } from 'lodash'
+import classNames from 'classnames'
+import { FileImportProps } from 'components/organisms/FileImport/FileImport'
+
+import classes from './styles.module.scss'
+
+type FileProps = Omit<FileImportProps, 'helperText' | 'fileButtonText'> & {
+  isDragAndDropOpen?: boolean
+  setFileContent?: Dispatch<SetStateAction<string>>
```
KaarelKa: **[blocker]**

Mentioned in previous CR as well, but these set state types for functions are too limiting. In the previous place we had to change it right away and for file input we won't be storing the files in state in the first cases, where we use it at all, possibly never will

src/components/organisms/DragAndDrop/DragAndDrop.tsx
```diff
@@ -0,0 +1,172 @@
...
+      >
+        {map(files, (file, index) => {
+          return (
+            <Fragment key={index}>
+              <File className={classes.icon} />
+              <div className={classes.fileDetailsContainer}>
+                <p className={classes.fileName}>{file?.name}</p>
+                <p className={classes.fileSize}>{formatFileSize(file?.size)}</p>
+              </div>
+              <Delete onClick={onClick} />
```
KaarelKa: **[blocker]**

1. onClick name here is quite confusing. For 99% of components we assume that if we pass the "onClick" prop to them, then this is a function which will be executed, when we click on that specific element. In this case onClick is actually passed through 2 components and used as an "onDelete" function.

2. Deletion should be handled inside this component mostly. We still need the "onDelete" callback for when a file is deleted, but the actual deletion (removal from the state) should be handled here

3. Everything clickable should still use the Button or BaseButton component

src/components/organisms/DragAndDrop/DragAndDrop.tsx
```diff
@@ -0,0 +1,172 @@
...
+          )
+        })}
+      </ul>
+      <p
+        hidden={!error || !files?.length}
+        className={classNames(files?.length && classes.errorText)}
+      >
+        {error}
+      </p>
+      <div className={classes.fileContent}>{fileContent}</div>
```
KaarelKa: **[blocker]**

We shouldn't show the content of the file here. For images this will be a page full of unreadable characters for example. Handling the file will be done outside of this component

src/components/organisms/DragAndDrop/DragAndDrop.tsx
```diff
@@ -0,0 +1,172 @@
...
+      </div>
+    </div>
+  )
+}
+
+const DragAndDrop: FC<FileProps> = ({
+  fileLabel,
+  onClick,
+  error,
+  kilobytesLabel,
```
KaarelKa: **[blocker]**

A lot of these texts could be internal ones. I doupt they will be different for different file inputs. Currently the component is quite hard to use, because of the amount of props we have to pass in

src/components/organisms/DragAndDrop/styles.module.scss
```diff
@@ -0,0 +1,98 @@
...
+@import 'styles/mixins';
+
+.dragAndDropContainer {
+  background: $color-text-button;
+  border: 1px solid $color-border-black-coral;
+  border-radius: 5px;
+  width: 318px;
+  padding: spacing(4) spacing(4) spacing(2) spacing(4);
+  position: relative;
+  &::after {
```
KaarelKa, line 20: **[non-blocker]**

Would be good to have some shadow around the triangle as well, otherwise it will not be visible, if the file input is on a white background

src/components/organisms/FileImport/FileImport.tsx
```diff
@@ -0,0 +1,92 @@
...
+  error?: string
+  dropFilesText?: string
+  dragFilesText?: string
+  dropFilesButtonText?: string
+  multipleFilesText?: string
+  dropButtonText?: string
+}
+
+const FileImport: FC<FileImportProps> = ({
+  helperText,
```
KaarelKa, line 26: **[blocker]**

Too many props make the component really hard to reuse. A lot of these are internal strings or button styling related props, which don't need to be changeable

KaarelKa (changes requested):

PR #34, opened 2023-06-21
249

IngaSokman:


src/components/organisms/modals/ModalRoot.tsx
```diff
@@ -1,26 +1,31 @@
...
 
 export enum ModalTypes {
   InstitutionSelect = 'institutionSelect',
+  Tooltip = 'tooltip',
 }
 
 // Add other modal props types here as well
-type ModalPropTypes = Omit<InstitutionSelectModalProps, 'closeModal'>
+type ModalPropTypes = Omit<InstitutionSelectModalProps, 'closeModal'> &
+  Omit<TooltipModalProps, 'closeModal'>
```
VaabelMaret: Does this ModalPropType have to take all the time every modals props?

src/components/organisms/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,44 @@
...
+    })
+  }
+
+  return (
+    <BaseButton onClick={handleModalOpen}>
+      <Icon icon={icon} ariaLabel={ariaLabel} />
+    </BaseButton>
+  )
+}
+
```
VaabelMaret, line 56: If I am correct that this tooltip modal is not working without importing and adding the 'ModalRoot' here.

src/components/organisms/modals/TooltipModal/TooltipModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+import classes from './styles.module.scss'
+
+export interface TooltipModalProps {
+  title?: string
+  textButtonContent?: string
+  modalContent?: string
+  href?: string
+  isModalOpen?: boolean
+  closeModal: () => void
+  onClose?: () => void
```
VaabelMaret: Where do you use onClose func? I don't see that you pass on any func to onClose.

VaabelMaret (changes requested):

src/components/organisms/Tooltip/Tooltip.tsx
IngaSokman: No, TooltipModal is added to ModalRoot so that Tooltip component can use TooltipModal as a function in here:
 `const handleModalOpen = () => {
    showModal(ModalTypes.Tooltip, {
      title: title,
      textButtonContent: textButtonContent,
      modalContent: modalContent,
      href: href,
    })
  }`

src/components/organisms/modals/TooltipModal/TooltipModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+        {
+          appearance: AppearanceTypes.Text,
+          children: textButtonContent,
+          size: SizeTypes.M,
+          href: href,
+          iconPositioning: IconPositioningTypes.Left,
+        },
+        {
+          appearance: AppearanceTypes.Secondary,
+          onClick: handleClose,
```
VaabelMaret: How about passing the closeModal func here or is this useCallback needed?

IngaSokman: I think it would be good to use useCallback here to prevent unnecessary re-renders.

IngaSokman: Double checked the useCallback, it does not make re-renders, removed it

VaabelMaret (approved):

src/components/organisms/modals/TooltipModal/TooltipModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+  textButtonContent?: string
+  modalContent?: string
+  href?: string
+  isModalOpen?: boolean
+  closeModal: () => void
+}
+
+const TooltipModal: FC<TooltipModalProps> = ({
+  title,
+  textButtonContent,
```
KaarelKa, line 25: **[non-blocker]**

Right now there is no point to do it, but once we add the page for the FAQ or (Kasutajajuhend) page, we should also add default values for textButtonContent and href, so it would navigate to the FAQ page by default

src/components/organisms/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,44 @@
+import { FC, SVGProps } from 'react'
+import BaseButton from 'components/atoms/BaseButton/BaseButton'
+import { showModal, ModalTypes } from 'components/organisms/modals/ModalRoot'
+import { IconProps } from 'components/molecules/Button/Button'
+
+interface TooltipProps {
+  icon?: FC<SVGProps<SVGSVGElement>>
```
KaarelKa, line 11: **[blocker]**

Let's add the default icon here. In most cases we will use the questionmark icon for this, so let's use that as the default.

KaarelKa (changes requested):

KaarelKa (approved):

PR #35, opened 2023-06-26
292

IngaSokman:


VaabelMaret (approved):

src/components/organisms/RolesTabs/RolesTabs.tsx
```diff
@@ -81,6 +81,9 @@ const RolesTabs: FC = () => {
     return <div />
   }
 
+  const filteredMainUsers = filter(existingRoles, { is_root: true })
```
KaarelKa: **[non-blocker]**

Naming is a bit confusing here. We are dealing with roles not users. Right now it's not much of a problem, but could be in the future, when we have added the pages for manipulating users. Will make it slightly harder to find places in the codebase where we are dealing with users

src/components/organisms/RolesTabs/RolesTabs.tsx
```diff
@@ -108,6 +114,7 @@ const RolesTabs: FC = () => {
             {...role}
             temporaryName={tabNames[role.id]}
             allPrivileges={allPrivileges}
+            isMainUser={isMainUser}
```
KaarelKa: **[blocker]**

Currently it seems that what you are checking whether the role has "is_root" = true or not. Current approach is 

1. Pick all roles that have is_root="true"
2. Create an array of all their id's
3. While mapping over all roles, check if their id is in that list

This seems to be a bit overengineered.
We are already mapping over all roles, when rendering the list of <RoleForm> components.
So the part we spread there (...role) also includes the "is_root" value for that form. I would suggest using the value from there directly, instead of passing a duplicate of the same value. (google the KISS principle) - Essentially do things in as simple a way as possible.

KaarelKa (changes requested):

src/components/organisms/RolesTabs/RolesTabs.tsx
IngaSokman: Good point yes, changed it

KaarelKa (approved):

PR #36, opened 2023-06-27
136

IngaSokman:


src/components/organisms/modals/DeleteRoleModal/DeleteRoleModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+} from 'components/organisms/ModalBase/ModalBase'
+
+export interface DeleteRoleModalProps {
+  title?: string
+  cancelButtonContent?: string
+  proceedButtonContent?: string
+  modalContent?: string
+  isModalOpen?: boolean
+  closeModal: () => void
+  onClose?: () => void
```
VaabelMaret: You don't pass the onClose func. What is the purpose of this function?

src/components/organisms/modals/ModalRoot.tsx
```diff
@@ -1,26 +1,32 @@
...
+
 export enum ModalTypes {
   InstitutionSelect = 'institutionSelect',
+  DeleteRole = 'deleteRole',
 }
 
 // Add other modal props types here as well
-type ModalPropTypes = Omit<InstitutionSelectModalProps, 'closeModal'>
+type ModalPropTypes = Omit<InstitutionSelectModalProps, 'closeModal'> &
+  Omit<DeleteRoleModalProps, 'closeModal'>
```
VaabelMaret: I think "or" is better than "and" here. We don't need to include all modal props all the time

VaabelMaret (changes requested):

src/components/organisms/modals/DeleteRoleModal/DeleteRoleModal.tsx
IngaSokman: Removed redundant code

src/components/organisms/modals/DeleteRoleModal/DeleteRoleModal.tsx
```diff
@@ -0,0 +1,57 @@
+import { FC } from 'react'
+import { AppearanceTypes, SizeTypes } from 'components/molecules/Button/Button'
+import ModalBase, {
+  ButtonPositionTypes,
+  TitleFontTypes,
+} from 'components/organisms/ModalBase/ModalBase'
+
+export interface DeleteRoleModalProps {
+  title?: string
+  cancelButtonContent?: string
```
KaarelKa: **[blocker]**

Either the name of the modal is not correct, or we are passing in way too many props:

1. If this is a modal just for deleting roles (as the name suggest):

 I don't think we need to have the "cancelButtonContent", "proceedButtonContent", "title", "handleProceed" props. Instead I would just pass the id of the role that we are deleting to the modal. This way it would be far easier to reuse it if we ever have other places, where we need to delete roles.

The "modalContent" prop seems to be something that is never passed, so there it should be removed, unless we know of some case in the future, where it will be passed.


2. OR, if this is suppose to be the Modal that we use for confirming any kind of deletion (roles, users, etc.), then these props should be fine, however the name of the component is misleading

KaarelKa (changes requested):

src/components/organisms/modals/ModalRoot.tsx
```diff
@@ -1,4 +1,5 @@
 import { InstitutionSelectModalProps } from './InstitutionSelectModal/InstitutionSelectModal'
```
KaarelKa, line 1: **[blocker]**

Typescript error in this file

src/components/organisms/modals/DeleteRoleModal/DeleteRoleModal.tsx
IngaSokman: Yes, it is used in other user management as well, forgot to change the name in here.

VaabelMaret (approved):

PR #37, opened 2023-06-29
266 progress bar

VaabelMaret:


IngaSokman:
I think we should make a new modal that is using the progressbar with cancel and proceed button.

VaabelMaret:
@IngaSokman  As far as I know there is another ticket for the modal. This ticket only contained a progress bar.

IngaSokman:
Ok, got it. In that case approved. Only merge conflicts need to be solved then.

IngaSokman (approved):

src/components/atoms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,38 @@
+import { FC } from 'react'
+import classes from './styles.module.scss'
```
KaarelKa: **[blocker]**

Let's change this to "classes.module.scss" before merging

src/components/atoms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,38 @@
...
+  isCurrentStep: boolean
+}
+
+interface ProgressBarProps {
+  steps: Step[]
+}
+
+const ProgressBar: FC<ProgressBarProps> = ({ steps }) => {
+  return (
+    <div className={classes.progressBarContainer}>
```
KaarelKa: **[blocker]** (Potentially)

Do we need this div around the ul ? Did a quick test here and it seems to be unnecessary

src/components/atoms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,38 @@
...
+          const { label, isCurrentStep } = step || {}
+
+          return (
+            <li
+              key={index}
+              className={classNames(classes.step, {
+                [classes.currentStep]: isCurrentStep,
+              })}
+            >
+              <p className={classes.stepNumber}>{index + 1}</p>
```
KaarelKa: **[non-blocker]**

span would be a more appropriate html tag, at least for the step number part

src/pages/Components/Components.tsx
```diff
@@ -214,6 +215,14 @@ const Test: FC = () => {
           iconPositioning={IconPositioningTypes.Right}
         />
       </DynamicForm>
+      <ProgressBar
```
KaarelKa, line 218: **[non-blocker]**

Not too important and you can decide this, when creating the actual modal, where this component is used, but gut feeling is that it would be more appropriate to use something like "activeStep" prop here. As far as I know it shouldn't be possible to have multiple active steps + it will probably be easier to use if we just change the "activeStep" prop, instead of mapping over the list of all steps every time we want to change the active step.

KaarelKa (changes requested):

KaarelKa (approved):

PR #39, opened 2023-07-04
148

IngaSokman:


src/pages/UserPage/UserPage.tsx
```diff
@@ -29,6 +35,23 @@ const UserPage: FC = () => {
...
   const userNameString = `${user.user.forename} ${user.user.surname}`
 
+  const handleArchiveModal = () => {
+    const isMainUser = some(user?.roles, (mainUser) => mainUser?.is_root)
+
+    !isMainUser &&
+      showModal(ModalTypes.UserAndRoleManagement, {
+        title: t('modal.archive_user_account'),
+        cancelButtonContent: t('button.no'),
+        proceedButtonContent: t('button.yes'),
```
VaabelMaret: [non-blocker]
Currently I have seen that in user and role management modals we have yes and no buttons. Do we have anything else as well? If there is mostly yes and no then these should be default values.

src/components/organisms/modals/UserAndRoleManagementModal/UserAndRoleManagementModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+  TitleFontTypes,
+} from 'components/organisms/ModalBase/ModalBase'
+
+export interface UserAndRoleManagementModalProps {
+  title?: string
+  cancelButtonContent?: string
+  proceedButtonContent?: string
+  modalContent?: string | ReactElement
+  isModalOpen?: boolean
+  closeModal: () => void
```
VaabelMaret, line 14: Why this closeModal prop is here, if you don't send any func with it? Are you planing to use the closeModal func?

src/pages/UserPage/UserPage.tsx
```diff
@@ -29,6 +35,23 @@ const UserPage: FC = () => {
 
   const userNameString = `${user.user.forename} ${user.user.surname}`
 
+  const handleArchiveModal = () => {
+    const isMainUser = some(user?.roles, (mainUser) => mainUser?.is_root)
```
VaabelMaret: [non-blocker]
I you have only on key then you can use shorthand version: 'some(user?.roles, 'is_root')'

VaabelMaret (approved):

src/components/organisms/modals/UserAndRoleManagementModal/UserAndRoleManagementModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+  handleProceed?: () => void
+  className?: string
+}
+
+const UserAndRoleManagementModal: FC<UserAndRoleManagementModalProps> = ({
+  title,
+  cancelButtonContent,
```
KaarelKa: **[blocker]**

Tend to agree with Maret here, that it would be good to at least have "yes" and "no" as the default values for proceedButtonContent and cancelButtonContent. Didn't go over all of the design, but I'm not sure if we will have any cases, where they are anything other than "yes" and "no" for confirmation modals.

Especially if this is not a generic confirmation modal, but rather something that is just used for "UserAndRoleManagement" as the name suggests.

src/pages/UserPage/UserPage.tsx
KaarelKa: "mainUser" naming here is quite confusing as well. We are dealing with roles, not users.

src/pages/UserPage/UserPage.tsx
```diff
@@ -29,6 +35,23 @@ const UserPage: FC = () => {
...
+
+    !isMainUser &&
```

This behaviour does not match the behaviour described in WIKI. Currently we are blocking the archiving of the user, if he has some role with "is_root" true, but according to wiki we should allow these types of users to be archived, unless we would be left with 0 users with that "is_root" role, after archiving. 

-- https://github.com/keeleinstituut/tv-tolkevarav/wiki/5.-Autoriseerimine,--kasutajate-lisamine-s%C3%BCsteemi-(Kasutajate-ja-rollide-haldus)/5d3d608250967b6b0b33de08da0360ed9491d22c#5-archiving-the-account-added-2601

-- Some extra discussion about it in Slack -- https://pundar.slack.com/archives/C053FD55ATT/p1687952259793729

Basically the idea here is that we don't do any check on FE and just show an error Alert, if BE returns the error, that this account can't be archived

KaarelKa (changes requested):

src/pages/UserPage/UserPage.tsx
IngaSokman: Yes, there is mostly "yes" and "no", there is also "cancel" and "activate". Will add the default values.

src/pages/UserPage/UserPage.tsx
IngaSokman: Yes, currently it is blocked by BE task #394, waiting for deploy

IngaSokman:
Ready for review

KaarelKa (approved):

PR #45, opened 2023-07-07
104 

VaabelMaret:


src/components/organisms/DataTable/DataTable.tsx
```diff
@@ -55,33 +63,49 @@ const DataTable = <TData extends object>(
...
+  })
+
+  useEffect(() => {
+    if (onPaginationChange) {
+      onPaginationChange({
+        per_page: pagination.pageSize,
+        page: pagination.pageIndex + 1,
+      })
+    }
+  }, [pagination])
```
KaarelKa, line 91: **[blocker]**

Hook warning here, either add the dependency or disable rule for this line. Disabling is also fine, in many cases we don't want everything in the dependencies array, but depends on the case.

src/components/organisms/DataTable/DataTable.tsx
```diff
@@ -1,4 +1,11 @@
-import { CSSProperties, ReactElement, Ref, forwardRef, useState } from 'react'
```
KaarelKa, line 1: **[blocker]**

With the change to ApiClient (returning response.data instead of response.data.data) some places currently broke. Mostly you should also go over the mutation queries we do. (useUpdateRole, useCreateRole, useDeleteRole). The data there is now stored under ['roles', roleId] key. in the onSuccess callback we change the internal state of react-query, so we would have correct information there.

The current logic there assumes that the correct data for a user is the "data" returned from ApiClient, which used to be the case. However the "data" is no longer the same, so it is currently not updating the internal state correctly

KaarelKa (changes requested):

src/components/organisms/DataTable/DataTable.tsx
```diff
@@ -55,33 +63,49 @@ const DataTable = <TData extends object>(
...
+        page: pagination.pageIndex + 1,
+      })
+    }
+  }, [pagination])
 
   const table = useReactTable<TData>({
     data,
     columns,
+    manualPagination: !!paginationData, // Tell react-table that you will handle the pagination manually
+    pageCount: total, // Provide the total number of pages
```
KaarelKa: **[blocker]**

"total" from BE tells you the total number of items that exist in this list, not the total number of pages. Currently AddedUsersTable thinks we have 2 pages of information and navigating to the second page means that the table disappears.

**Extra**: Would also be good to not hide the entire table, when we have no results, but we know that there should be more than 1 page of information. Will help us avoid some edge cases in the future. (Cases where we have 2 pages of information when rendering, some information is deleted from Database and user will end up with no table, when they try to go to the next page)

KaarelKa (changes requested):

IngaSokman:
[blocker] I think the sorting svg icon is a bit smaller than on the design.

src/components/organisms/TableHeaderGroup/TableHeaderGroup.tsx
```diff
@@ -96,8 +95,8 @@ const HeaderItem = <TData extends object>({
   const handleOnFiltering = (value: string | string[]) => {
     const filterKey = keys(filterOption)[0]
 
-    if (onColumnFiltersChange) {
-      onColumnFiltersChange({ [filterKey]: value })
+    if (onFiltersChange) {
+      onFiltersChange({ [filterKey]: value })
     }
   }
```
IngaSokman, line 102: [blocker] I think the sorting svg icon is a bit small than on the design

IngaSokman (commented):
Noticed that when sorting button is being used, the roles column disappears, should it behave that way?

KaarelKa (approved):

PR #50, opened 2023-07-16
New-267-added modal with progressbar

VaabelMaret:


IngaSokman (approved):

src/components/organisms/modals/FormProgressModal/FormProgressModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+export interface FormProgressProps {
+  steps?: Step[]
+  formData?: FormDataProps[]
+  isModalOpen?: boolean
+  closeModal: () => void
+}
+
+const FormProgressModal: FC<FormProgressProps> = ({
+  steps,
+  formData,
```
KaarelKa, line 25: **[non-blocker]**

Not necessary, but a less error-prone data structure here would be to combine "formData" and "steps" props here.
Basically sth like


`
steps = [
    {
        label: 'Step1',
        title: 'Title1',
        helperText: 'Helper1',
        content: <div>Content1>
    },
    ...
]
`

This way it will be very clear what step matches what content + helps us avoid cases, where the size of "steps" and "formData" arrays don't match.

KaarelKa (approved):

src/components/organisms/modals/FormProgressModal/FormProgressModal.tsx
VaabelMaret: Good Idea

KaarelKa:
@VaabelMaret You can merge this, after merge conflicts are resolved

PR #52, opened 2023-07-18
140

IngaSokman:


src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -40,28 +40,33 @@ const DatePickerComponent = ({
   onChange,
   ...rest
 }: DatePickerComponentProps) => {
-  const handleDateChange: ReactDatePickerProps['onChange'] = (value) =>
-    onChange(changeDateToString(value))
+  const handleDateChange: ReactDatePickerProps['onChange'] = (value) => {
+    return onChange(changeDateToString(value))
```
KaarelKa: **[non-blocker]**

No need to return the onChange function here, you can just call it

src/helpers/index.ts
```diff
@@ -87,6 +87,18 @@ export const objectsToCsvFile = <ValuesType>(
   return file
 }
 
+export const formatDate = (
```
KaarelKa: **[blocker]**

This looks like a custom implementation of the dayjs.format method. Would prefer to use dayjs for anything related to dates.

For example the use case you have in DatePickerInput:

1. You have value in format DD/MM/YYYY
2. You want to convert it to YYYY-MM-DD format
3. You want to add 1 year to current date and get the Date object from it

Steps to achieve this would be:

1. Convert your current value to a dayjs object:

`const valueAsDayjs = dayjs(value, 'DD/MM/YYYY')`

**Note:** Dayjs doesn't support the DD/MM/YYYY format parsing by default, so we need to extend it

`import customParseFormat from 'dayjs/plugin/customParseFormat'`
`dayjs.extend(customParseFormat)`

2. Convert that object into the format you need:

`valueAsDayjs.format('YYYY-MM-DD')`

3. Add + 1 year to current date and get the date object

`dayjs().add(1, 'year').toDate()`


Basically we should be able to do anything we need with dates and times with dayjs and there shouldn't really be any need for implementing the same or similar features manually

src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -40,28 +40,33 @@ const DatePickerComponent = ({
...
         id={name}
-        selected={value ? new Date(formattedDayValue) : null}
+        selected={value ? new Date(splittedDayValue) : null}
         dateFormat={'dd.MM.yyyy'}
         locale="et-EE"
         placeholderText={placeholder}
         aria-label={ariaLabel}
         disabled={disabled}
+        minDate={new Date()}
+        maxDate={new Date(formattedDate)}
```

Might have missed it, but I don't think we should have a min and max date limit for every date input, unless it is defined like this somewhere. For example we will have Order deadlines, which shouldn't be limited to max 1 year, there could be Orders that take multiple years. We also might have some date of birth input at some point, which would require picking dates from the past.

These are just 2 examples I can think of right now, but there might be others, so we shouldn't add this sort of a limitation to all our DatePicker components.

These should be optional props that can be passed in where needed

src/pages/UserPage/UserPage.tsx
```diff
@@ -10,20 +14,67 @@ import { useTranslation } from 'react-i18next'
...
   })
   const { archiveUser, isLoading: isArchiving } = useArchiveUser({
     userId: userId,
   })
-  const navigate = useNavigate()
+  const { deactivateUser, isLoading: isDeactivating } = useDeactivateUser()
+  const deactivationDate = user?.deactivation_date || ''
+
+  const currentFormattedDefaultDate = format(new Date(), 'dd/MM/yyyy')
+  const user_deactivation_date = currentFormattedDefaultDate
```

Let's use dayjs here as well for consistency (same goes for every place where we deal with dates).

**[non-blocker]**

Reassigning a variable to a new variable seems unnecessary here 

const user_deactivation_date = currentFormattedDefaultDate

I would rather skip it. If you prefer the name user_deactivation_date, then you can name it that right away, no need to create some new variable in the middle

src/pages/UserPage/UserPage.tsx
```diff
@@ -48,19 +111,65 @@ const UserPage: FC = () => {
       },
     })
   }
+  const handleDeactivateModal = () => {
+    showModal(ModalTypes.UserAndRoleManagement, {
+      title: t('modal.deactivate_user'),
+      cancelButtonContent: t('button.cancel'),
+      proceedButtonContent: t('button.yes'),
```

default value for proceedButtonContent is already "button.yes", so this prop could be skipped, but not a big problem, it can stay as well

src/pages/UserPage/UserPage.tsx
```diff
@@ -48,19 +111,65 @@ const UserPage: FC = () => {
...
+        deactivateUser({
+          user_deactivation_date: values?.user_deactivation_date || '',
+          userId: userId || '',
+        })
+      ),
+      dynamicForm: <DynamicForm fields={fields} control={control} />,
+    })
+  }
+
+  const handleEditModal = () => {
```

These 2 modals seems to be identical, except for the title. Would be better to avoid code repetition here and just make the "title" dynamic, based on which modal will be opened. Basically you could just pass the correct title to handleDeactivateModal method, or pass some boolean "isEdit: true/false" to decide on which title to show

KaarelKa:
**[blocker]**

1. We should also show success notification, whenever some user action is successful. Doesn't seem to be mentioned in wiki, but let's still add it. So in this case we need to add it for both deactivation request success and changing the deactivation date success. 

2. It seems that currently the data on the page is not updated, when user sets a deactivation date or changes it. It only happens, once the page is refreshed or user details are refetched. Check the useUpdateUser hook for an example how to do this (more examples in the useRoles file). Basicallt what is happening there is that when the request is successful, we change the internal state inside react-query based on the response. (Similar to what we do in redux reduces, when a request is successful or fails).

KaarelKa (changes requested):

src/pages/UserPage/UserPage.tsx
```diff
@@ -49,18 +151,46 @@ const UserPage: FC = () => {
...
+      title: title,
+      cancelButtonContent: t('button.cancel'),
+      modalContent: t('modal.deactivate_user_content'),
+      className: classes.deactivateContent,
+      handleProceed: handleSubmit(onSubmit),
+      dynamicForm: <DynamicForm fields={fields} control={control} />,
+    })
+  }
+
+  const currentFormattedDate = dayjs(new Date()).format('YYYY-MM-DD')
```
KaarelKa: **[non-bocker]**

dayjs() does the same thing as dayjs(new Date()), so the "new Date()" part is unnecessary with dayjs

src/pages/UserPage/UserPage.tsx
```diff
@@ -34,12 +133,15 @@ const UserPage: FC = () => {
   }
 
   const userNameString = `${user.user.forename} ${user.user.surname}`
+  const isUserDeactivated = !!deactivationDate
+
+  const deactivatedDate = dayjs(deactivationDate, 'YYYY-MM-DD').toDate()
+  const isDeactivationDatePastCurrentDate =
```

Would be better to use dayjs whereever possible and basically use Date only when there are no other options.
Currently dayjs is used, but not for comparing 2 dates, which is also a feature that is build into dayjs.

So if you want to check whether 1 date is before or after another, you can use dayjs objects instead and use the isBefore or isAfter method of dayjs.

Could reduce the code a bit + avoid unnecessary conversions to Date object.

https://day.js.org/docs/en/query/is-after
https://day.js.org/docs/en/query/is-before#docsNav

KaarelKa (approved):

PR #53, opened 2023-07-19
146

IngaSokman:


src/pages/UserPage/UserPage.tsx
```diff
@@ -91,7 +97,44 @@ const UserPage: FC = () => {
...
   ]
 
-  const onSubmit: SubmitHandler<FormValues> = useCallback(
+  const roleOptions = compact(
+    map(existingRoles, ({ name, id }) => {
+      if (name) {
+        return {
+          label: name,
+          value: name,
+          id: id,
```

I don't think we need this id field. "value" field is meant for exactly this purpose. Basically for dropdowns "value" is what we need from code side, in this case what we need for sending information to BE and "label" is only for display purposes. Currently you are using "id" instead of "value" and essentially not using "value" at all

src/pages/UserPage/UserPage.tsx
```diff
@@ -124,6 +167,41 @@ const UserPage: FC = () => {
     [deactivationDate, userId, deactivateUser, t, name, setError]
   )
 
+  const onActivateSubmit: SubmitHandler<FormValues> = useCallback(
+    async (values) => {
+      const { user_activation_notification } = values
+
+      const roleIdsArray = roleOptions
```
KaarelKa: **[followup]**

As a followup to the previous comment, if you would use just the "value" for those options, there would be no need to filter or map anything here. "user_activation_notification" would already be an array of role id-s and could be passed directly to BE.

src/pages/UserPage/UserPage.tsx
```diff
@@ -124,6 +167,41 @@ const UserPage: FC = () => {
...
+        .filter((obj) => values?.user_activation_roles?.includes(obj.label))
+        .map((obj) => obj?.id)
+
+      const payload = {
+        userId: userId || '',
+        notify_user: user_activation_notification || false,
+        roles: roleIdsArray || [],
```

It would be good to name the input fields with the same names that BE expects (so in this case it would be institution_user_id, notify_user, roles) instead of (userId, user_activation_notification, user_activation_roles).

Not strictly necessary, just we can avoid a lot of renaming of variables this way + will be easier to search for the correct places in the codebase, if needed.

src/hooks/requests/useUsers.ts
```diff
@@ -198,3 +198,44 @@ export const useDeactivateUser = ({ userId }: { userId?: string }) => {
     isLoading,
   }
 }
+
+export const useActivateUser = ({ userId }: { userId?: string }) => {
```
KaarelKa: **[blocker]**

Didn't notice before, but this also applies for useDeactivateUser:

Currently using this hook means that we need to pass the userId in 2 different places, which can be really errorprone + makes this less convenient to use. Since we need the userId in the constructor of the hook, we can remove it from the "mutationFn" arguments

src/pages/UserPage/classes.module.scss
```diff
@@ -39,18 +39,35 @@
...
+}
+
+.activateContent h1,
 .deactivateContent h1,
+.activateContent > div > div,
 .deactivateContent > div > div {
   max-width: 70%;
 }
+
+.activateContent > div > div > form {
```
KaarelKa: **[blocker]**

Might not be able to get rid of all of these, but let's try to reduce the number of these types of style overwrites where possible. They work, but they are also really error prone, since they depend on the exact structure of html and will break, when that structure changes even a bit.

For example you could pass className prop to the DynamicForm directly, instead of the ".activateContent > div > div > form " targeting

Didnät go over the other options here, but I'm guessing you could get rid of some others as well. (For example the :first-child selector below seems to be targeting one of the inputs inside the form, so we should instead pass a custom className to that input)

KaarelKa (changes requested):

KaarelKa (approved):

PR #54, opened 2023-07-19
182 institution view and 211 endpoints and updating data

VaabelMaret:


IngaSokman (changes requested):
**[blocker]** When the user presses the "edit" button, the question mark icon moves to the right.

IngaSokman:
**[blocker]** On the wiki and design it says "Salvesta" not "Salvesta muudatused" for the save button.

IngaSokman:
**[blocker]** When I update the institution name, save it and then want to change the institution, the new changed name does not appear on the institution options.

src/components/organisms/forms/InstitutionForm/InstitutionForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+import classNames from 'classnames'
+
+interface FormValues {
+  name: string
+  short_name?: string | null
+  email?: string | null
+  phone?: string | null
+}
+
+type InstitutionFormProps = InstitutionType
```
IngaSokman: **[non-blocker]** You could add the InstitutionType directly to InstitutionForm

IngaSokman (changes requested):

src/components/molecules/cheatSheets/InstitutionManagementCheatSheet.tsx
```diff
@@ -0,0 +1,22 @@
+import { useTranslation } from 'react-i18next'
```
KaarelKa, line 1: **[non-blocker]**

@IngaSokman @VaabelMaret For future reference, we will be skipping these cheat sheets for now. They will be added with a slightly different structure

But no need to remove it.

src/components/organisms/forms/InstitutionForm/InstitutionForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+  })
+
+  const fields: FieldProps<FormValues>[] = [
+    {
+      inputType: InputTypes.Text,
+      ariaLabel: t('label.institution'),
+      label: t('label.institution'),
+      disabled: !isUpdatingData,
+      name: 'name',
+      className: classNames(classes.inputInternalPosition, {
```
KaarelKa, line 72: **[non-blocker]**

Let's leave this for now, but we might change it later. We have the DisplayValue component which will be displayed instead of the actual input, when we pass "onlyDisplay: true" as a prop here. It was meant for similar usecases, however it doesn't show the label currently, so would need to be modified.

If we will deal with any future places, where we need this again, then it might make sense to do the modification for the DisplayValue component, but not right now

KaarelKa (approved):

PR #58, opened 2023-07-20
New-148-improvements Add translations, add success notifications, dis…

IngaSokman:
…able archive button when user archived

### DIFF
--- a/src/hooks/requests/useUsers.ts
+++ b/src/hooks/requests/useUsers.ts
@@ -147,7 +147,9 @@ export const useArchiveUser = ({
   const { mutateAsync: archiveUser, isLoading } = useMutation({
     mutationKey: ['users', institution_user_id],
     mutationFn: async () =>
-      apiClient.post(endpoints.ARCHIVE_USER, institution_user_id),
+      apiClient.post(endpoints.ARCHIVE_USER, {
+        institution_user_id: institution_user_id,
+      }),
 
     onSuccess: ({ data }) => {
       queryClient.setQueryData(
--- a/src/hooks/requests/useUsers.ts
+++ b/src/hooks/requests/useUsers.ts
@@ -138,13 +138,33 @@ export const useDownloadUsers = () => {
   }
 }
 
-export const useArchiveUser = ({ userId }: { userId?: string }) => {
-  const { mutate: archiveUser, isLoading } = useMutation({
-    mutationKey: ['users', userId],
-    mutationFn: () => {
-      return apiClient.post(endpoints.ARCHIVE_USER, {
-        institution_user_id: userId,
-      })
+export const useArchiveUser = ({
+  institution_user_id,
+}: {
+  institution_user_id?: string
+}) => {
+  const queryClient = useQueryClient()
+  const { mutateAsync: archiveUser, isLoading } = useMutation({
+    mutationKey: ['users', institution_user_id],
+    mutationFn: async (payload: UserStatusType) =>
+      apiClient.post(endpoints.ARCHIVE_USER, {
+        ...payload,
+      }),
+
+    onSuccess: ({ data }) => {
+      queryClient.setQueryData(
+        ['users', institution_user_id],
+        // TODO: possibly will start storing all arrays as objects
+        // if we do, then this should be rewritten
+        (oldData?: UsersDataType) => {
+          const { data: previousData } = oldData || {}
+
+          if (!previousData) return oldData
+          const newData = { ...oldData, ...data }
+
+          return { data: newData }
+        }
+      )
     },
   })
 
@@ -204,11 +224,10 @@ export const useActivateUser = ({
   const queryClient = useQueryClient()
   const { mutateAsync: activateUser, isLoading } = useMutation({
     mutationKey: ['users', institution_user_id],
-    mutationFn: async (payload: UserStatusType) => {
-      return apiClient.post(endpoints.ACTIVATE_USER, {
+    mutationFn: async (payload: UserStatusType) =>
+      apiClient.post(endpoints.ACTIVATE_USER, {
         ...payload,
-      })
-    },
+      }),
 
     onSuccess: ({ data }) => {
       queryClient.setQueryData(
--- a/src/hooks/requests/useUsers.ts
+++ b/src/hooks/requests/useUsers.ts
@@ -146,10 +146,8 @@ export const useArchiveUser = ({
   const queryClient = useQueryClient()
   const { mutateAsync: archiveUser, isLoading } = useMutation({
     mutationKey: ['users', institution_user_id],
-    mutationFn: async (payload: UserStatusType) =>
-      apiClient.post(endpoints.ARCHIVE_USER, {
-        ...payload,
-      }),
+    mutationFn: async () =>
+      apiClient.post(endpoints.ARCHIVE_USER, institution_user_id),
 
     onSuccess: ({ data }) => {
       queryClient.setQueryData(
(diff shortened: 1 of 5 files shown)


src/pages/UserPage/UserPage.tsx
```diff
@@ -52,20 +52,20 @@ const UserPage: FC = () => {
...
   const { deactivateUser, isLoading: isDeactivating } = useDeactivateUser({
     institution_user_id: userId,
   })
   const deactivationDate = user?.deactivation_date || ''
-  const forename = user?.user?.forename || ''
-  const surname = user?.user?.surname || ''
-  const name = `${forename} ${surname}`
+  const name = `${user?.user.forename} ${user?.user.surname}`
+
+  const isUserArchived = user?.archived_at !== null
```
KaarelKa, line 68: **[non-blocker]**

User actually also has the "status" field, which might be more reliable than checking whether "archived_at" date exists or not, but currently I'm not aware of any cases, where a user who isn't archived can have the "archived_at" value filled. Possibly in the future, when there will be some way to unarchive an account, but no need to deal with this now

src/pages/UserPage/UserPage.tsx
```diff
@@ -212,10 +244,7 @@ const UserPage: FC = () => {
       title: t('modal.archive_user_account'),
       modalContent: t('modal.archive_user_content'),
       className: classes.archiveContent,
-      handleProceed: () => {
-        archiveUser()
-        navigate('/settings/users')
-      },
+      handleProceed: handleSubmit(onArhiveSubmit),
```

We shouldn't use "handleSubmit" here, since we are not submitting any form

src/pages/UserPage/UserPage.tsx
```diff
@@ -130,13 +130,38 @@ const UserPage: FC = () => {
...
   ]
 
+  const onArhiveSubmit: SubmitHandler<FormValues> = useCallback(
+    async (values) => {
+      const payload: UserStatusType = { ...values, institution_user_id }
+
+      navigate('/settings/users')
+
+      try {
+        await archiveUser(payload)
```

As far as I know we shouldn't send the values from activate/deactivate modal forms into the archiveUser endpoint. There should be no need for it and can cause some nasty sideeffects

src/pages/UserPage/UserPage.tsx
```diff
@@ -130,13 +130,38 @@ const UserPage: FC = () => {
...
+
+        showNotification({
+          type: NotificationTypes.Success,
+          title: t('notification.announcement'),
+          content: t('success.user_archived', { name }),
+        })
+      } catch (errorData) {
+        const typedErrorData = errorData as ValidationError
```

No need to handle validation errors we get from archiving endpoint just to show the errors in deactivate/activate modal. Can cause some very weird behaviour where we get errors from archive user endpoint because we are passing fields there that it doesnt accept and set them as errors for the "activate" modal form. When the user then opens the "activate/deactivate" user modal, they will see some errors that don't make any sense + archiving can potentially fail because of these extra parameters that we send

src/pages/UserPage/UserPage.tsx
```diff
@@ -168,7 +193,15 @@ const UserPage: FC = () => {
 
   const onActivateSubmit: SubmitHandler<FormValues> = useCallback(
     async (values) => {
-      const payload: UserStatusType = { ...values, institution_user_id }
+      const { notify_user } = values
+
+      const payload: UserStatusType = {
+        ...values,
+        institution_user_id,
+        ...(notify_user === undefined
```
KaarelKa: **[non-blocker]** Could be written as notify_user: !!notify_user

src/pages/UserPage/UserPage.tsx
```diff
@@ -130,13 +130,38 @@ const UserPage: FC = () => {
...
       label: t('label.user_activation_notification'),
       ariaLabel: t('label.user_activation_notification'),
-      rules: {
-        required: true,
-      },
       className: classes.checkboxInputClass,
     },
   ]
 
+  const onArhiveSubmit: SubmitHandler<FormValues> = useCallback(
```
KaarelKa: **[non-blocker]** "onArchiveSubmit" :D

KaarelKa (changes requested):

src/hooks/requests/useUsers.ts
```diff
@@ -138,13 +138,33 @@ export const useDownloadUsers = () => {
...
-      })
+export const useArchiveUser = ({
+  institution_user_id,
+}: {
+  institution_user_id?: string
+}) => {
+  const queryClient = useQueryClient()
+  const { mutateAsync: archiveUser, isLoading } = useMutation({
+    mutationKey: ['users', institution_user_id],
+    mutationFn: async (payload: UserStatusType) =>
```
KaarelKa: **[blocker]**

There should be no need for payload here. As far as I know, we only need to send institution_user_id to the archive user endpoint and this is already passed as an argument to the hooks, so no need to pass it again

src/pages/UserPage/UserPage.tsx
```diff
@@ -130,13 +130,25 @@ const UserPage: FC = () => {
...
       label: t('label.user_activation_notification'),
       ariaLabel: t('label.user_activation_notification'),
-      rules: {
-        required: true,
-      },
       className: classes.checkboxInputClass,
     },
   ]
 
+  const onArchiveSubmit: SubmitHandler<FormValues> = useCallback(async () => {
```

Type here seems to be wrong. This is the type for a submit function of a form, however we are not submitting any form with this function. Same issue with naming. the naming suggest we are submitting a form.

src/pages/UserPage/UserPage.tsx
```diff
@@ -130,13 +130,25 @@ const UserPage: FC = () => {
...
+    const payload: UserStatusType = { institution_user_id }
```

Mentioned in the comment for the hook. No need to pass institution_user_id from here

src/pages/UserPage/UserPage.tsx
```diff
@@ -130,13 +130,25 @@ const UserPage: FC = () => {
...
+    navigate('/settings/users')
```

We shouldn't navigate away from the screen before we know whether the user archiving was successful or not.

src/pages/UserPage/UserPage.tsx
```diff
@@ -130,13 +130,25 @@ const UserPage: FC = () => {
...
+
+    try {
+      await archiveUser(payload)
+
+      showNotification({
+        type: NotificationTypes.Success,
+        title: t('notification.announcement'),
+        content: t('success.user_archived', { name }),
+      })
+    } catch (_) {}
```
KaarelKa, line 147: **[non-blocker]**

No error displayed, when user archiving fails with some validation error. You can leave this for now, I can add it in the "cleanup" branch

src/pages/UserPage/UserPage.tsx
```diff
@@ -168,8 +180,13 @@ const UserPage: FC = () => {
 
   const onActivateSubmit: SubmitHandler<FormValues> = useCallback(
     async (values) => {
-      const payload: UserStatusType = { ...values, institution_user_id }
+      const { notify_user } = values
 
+      const payload: UserStatusType = {
+        ...values,
+        institution_user_id,
+        ...{ notify_user: !!notify_user },
```

Why is the spreading part here needed ? Could just be notify_user: !!notify_user

KaarelKa (changes requested):

src/hooks/requests/useUsers.ts
```diff
@@ -138,13 +138,31 @@ export const useDownloadUsers = () => {
...
+export const useArchiveUser = ({
+  institution_user_id,
+}: {
+  institution_user_id?: string
+}) => {
+  const queryClient = useQueryClient()
+  const { mutateAsync: archiveUser, isLoading } = useMutation({
+    mutationKey: ['users', institution_user_id],
+    mutationFn: async () =>
+      apiClient.post(endpoints.ARCHIVE_USER, institution_user_id),
```
KaarelKa: **[blocker]** (potentially)

Did you test this ? I don't think this will work, data structure seems to be wrong

KaarelKa (changes requested):

KaarelKa (approved):

PR #63, opened 2023-07-26
93

IngaSokman:


src/components/molecules/cheatSheets/TagManagementCheatSheet.tsx
```diff
@@ -0,0 +1,23 @@
+import { useTranslation } from 'react-i18next'
```
KaarelKa, line 1: **[non-blocker]**

For future reference, there is no need to add these for now, they will be stored in a different structure, but no need to remove this

src/hooks/requests/useTags.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export const useBulkCreate = () => {
+  const queryClient = useQueryClient()
+  const { mutateAsync: createTags, isLoading } = useMutation({
+    mutationKey: ['tags'],
+    mutationFn: async (payload: TagsType) =>
+      apiClient.post(endpoints.CREATE_TAGS, payload),
+
+    onSuccess: ({ tags }) => {
```
KaarelKa: **[blocker]**

Values stored incorrectly in state of react-query. This means that after adding the tags, they don't show up in the list, before we refetch all of them.

1. BE returns the new tags under "data" key, not "tags", so this "tags" will always be undefined

2. oldData here has wrong type. It is typed as "TagsType" so again with key "tags", however the actual oldData inside react-query state uses the "data" key here (Same way it is stored in useFetchTags)

3.  const newData = { ...oldData, ...tags } -- since tags is always undefined and oldData is just an object with the structure "{ data: Tags[] }" , the newData will just be exactly the same as oldData

4. As you are setting data for the "tags" key, the data needs to be in the same format as it was previously, which means the structure should be "{ data: Tags[] }". However you are returning the structure: "{ tags: { data: Tags[] } }"

src/hooks/useValidators.ts
```diff
@@ -40,11 +50,27 @@ const useValidators = () => {
     return true
   }
 
+  const tagInputValidator: Validate<
+    string | ObjectType | undefined,
+    FormValues
+  > = (value) => {
+    const inputValue = typeof value === 'string' ? value : ''
+
+    if (!inputValue || hasValueOver50Chars(inputValue))
```
KaarelKa: **[non-blocker]**

Not a big issue, but the first "if" statement here already catches the case, where inputValue is falsey, which means that the input length error will be displayed in that case. There is no need to check it in the other "if" cases

src/pages/Tags/Tags.tsx
```diff
@@ -1,9 +1,242 @@
...
+  const tagCategoryOptions = map(tags, ({ type }) => {
+    return {
+      label: type || '',
+      value: type || '',
+    }
+  })
+
+  const uniqueTagCategoryOptions = uniqBy(tagCategoryOptions, 'label')
+
+  const updatedDataWithoutSkillsObject = filter(
```
KaarelKa: **[blocker]**

We should have 4 categories according to the wiki, but with this logic we only show categories that have some tags in them, which also means that if a category has no tags, there is no way to add any tags to that category. Currently the only category that has tags is "Oskused", which can't be edited, so effectively there is no way to add any tags right now.

src/pages/Tags/Tags.tsx
```diff
@@ -1,9 +1,242 @@
...
+    reset,
+  } = useForm<FormValues>({
+    reValidateMode: 'onSubmit',
+  })
+
+  const tagFields: FieldProps<FormValues>[] = [
+    {
+      inputType: InputTypes.Text,
+      ariaLabel: t('tag.tag_name'),
+      label: 'Nimetus',
```
KaarelKa: **[blocker]**

All the text we show should come from BE, content management or translations. For label it should be translations. If we leave these hardcoded strings in only 1 language into the codebase, it will be really hard to find all of them later

src/pages/Tags/Tags.tsx
```diff
@@ -1,9 +1,242 @@
...
+  ]
+
+  const categoryFields: FieldProps<FormValues>[] = [
+    {
+      inputType: InputTypes.Selections,
+      name: 'tagCategorySelection',
+      ariaLabel: t('tag.select_tag_category'),
+      options: updatedDataWithoutSkillsObject,
+      placeholder: t('tag.select_tag_category'),
+      multiple: true,
```
KaarelKa: **[blocker]**

1. According to wiki this should be a single category input. Tags are added to 1 category at a time.
2. This means we also don't need the "buttons: true" here

src/pages/Tags/Tags.tsx
```diff
@@ -1,9 +1,242 @@
...
+            onClick={handleSubmit(onTagsSubmit)}
+            loading={isCreatingTags}
+            disabled={!areInputFieldsDirty}
+            hidden={!includes(userPrivileges, Privileges.AddTag)}
+          />
+        </div>
+      </Container>
+
+      <div className={classes.categoryContainer}>
+        {map(uniqueCategoryTypes, (type) => (
```
KaarelKa: **[non-blocker]**

There actually is no need to split uniqueCategoryTypes from groupedCategoryData for mapping this data.
you could instead use 

map(groupedCategoryData, (tags, type) => ...

instead and use the "type" the same way it is used now + replace groupedCategoryData[type] with "tags". 

Not a major issue, just would mean less code and less complexity

src/pages/Tags/Tags.tsx
```diff
@@ -1,9 +1,242 @@
...
+          <Container key={type} className={classes.category}>
+            <span className={classes.tagCategoryNameContainer}>
+              <div className={classes.categoryName}>{type}</div>
```
KaarelKa: **[blocker]**

1. Text should always be in the appropriate html tags. div is not meant for text. 
2. You are currently displaying the "type" value as text to the user, however this needs to be translatable. (Everything displayed to the user should be translatable).
3. You are using "span" tag as a container for a div + a button, however <span> tag is meant for text

src/pages/Tags/Tags.tsx
```diff
@@ -1,9 +1,242 @@
...
+              <Button
+                appearance={AppearanceTypes.Text}
+                icon={EditIcon}
+                className={classes.editIcon}
+                hidden={type === t('tag.skills_tag')}
```
KaarelKa: **[blocker]**

This is extremely error-prone and will cause an issue right away in any other language. You are checking a constant value (type) against a translatable value, meant only for displaying (t('tag.skills_tag')). This will break as soon as the user switched to another language, or the client decides to use some other word for displaying "tag.skills_tag" text somewhere

src/pages/Tags/classes.module.scss
```diff
@@ -0,0 +1,101 @@
...
+  margin-bottom: spacing(2);
+}
+
+.tagInputField input {
+  width: 508px;
+  height: 100%;
+  padding: spacing(2);
+}
+
+.newRow {
```
KaarelKa, line 52: **[blocker]**

seems to be unused

src/types/tags.ts
```diff
@@ -0,0 +1,16 @@
+export interface TagType {
+  id?: string
+  institution_id?: string
+  name?: string
+  type?: string
```
KaarelKa: **[blocker]**

type should not a string, it can only be 1 of 4 constant values

KaarelKa (changes requested):

src/hooks/requests/useTags.ts
VaabelMaret: [blocker]

This is not working. You have data variable inside new data object not tags

src/hooks/requests/useTags.ts
```diff
@@ -0,0 +1,49 @@
...
+        ['tags'],
+        // TODO: possibly will start storing all arrays as objects
+        // if we do, then this should be rewritten
+        (oldData?: TagsType) => {
+          const { tags: previousData } = oldData || {}
+
+          if (!previousData) return oldData
+          const newData = { ...oldData, ...tags }
+
+          return { tags: newData }
```
VaabelMaret: [blocker]

with this solution the tags array won't update automatically when you have added new tag

src/hooks/useValidators.ts
```diff
@@ -14,6 +16,14 @@ const picIsCorrect = (pic: string) =>
     pic
   )
 
+const hasValueOver50Chars = (tagInput: string) => tagInput?.length > 50
```
VaabelMaret, line 17: [non-blocker]

you have already imported lodash, why don't you use the lodash size here for consistency?

src/pages/Tags/classes.module.scss
```diff
@@ -0,0 +1,101 @@
+@import 'styles/variables';
+@import 'styles/mixins';
+
+.tagsHeader {
+  display: grid;
+  grid-template-columns: max-content auto max-content max-content;
```
VaabelMaret, line 6: [blocker]
 You only have two things in the header container, but you've split the content into four parts

`grid-template-columns: max-content max-content;`

src/pages/Tags/Tags.tsx
```diff
@@ -1,9 +1,242 @@
...
+          modalContent={<TagsCheatSheet />}
+        />
+      </div>
+      <Container className={classes.tagsContainer}>
+        <div>
+          <h4 className={classes.addingTag}>{t('tag.adding_tag')}</h4>
+          <p>{t('tag.naming_tag')}</p>
+        </div>
+
+        <div className={classes.addingTagsSeparator} />
```
VaabelMaret, line 172: [blocker]

no need for separator div, just add` border-right: 1px solid $color-background-dark;` for naming tag container or border-left for tagsSection div

src/pages/Tags/classes.module.scss
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.tagsContainer {
+  display: grid;
+  grid-template-columns: auto auto auto auto;
+  margin-top: spacing(6);
+  padding: spacing(6) spacing(8);
+}
+
+.tagsSection {
```
VaabelMaret, line 18: [non-blocker]

with display: grid you will have less code in here

src/pages/Tags/classes.module.scss
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.tagInputField input {
+  width: 508px;
+  height: 100%;
+  padding: spacing(2);
+}
+
+.newRow {
+  flex-shrink: 1;
```
VaabelMaret, line 53: [non-blocker]

don't understand why you need this style?

src/pages/Tags/classes.module.scss
```diff
@@ -0,0 +1,101 @@
...
+.categoryContainer {
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr;
+  padding-top: spacing(8);
+  grid-gap: spacing(6);
+}
+
+.category {
+  padding: spacing(2) spacing(4);
+  display: block;
```
VaabelMaret, line 69: [blocker]

no need to make this container display:block. You can use it as grid

VaabelMaret (changes requested):

PR #68, opened 2023-08-07
125

VaabelMaret:


IngaSokman (commented):
I think there is some weird behaviour: when I delete a bunch of tags in a category, the category containers will be switched for some time. Pictures below before deleting tags and after. 
Screenshot 2023-08-07 at 15 22 57 (2)
Screenshot 2023-08-07 at 15 23 22 (2)

IngaSokman (commented):
Just a sidenote, noticed that the error messages close button is in different color that in the design: 
Screenshot 2023-08-07 at 15 40 39

IngaSokman (commented):
Question about deleting tags. When the user deletes some tags on modal, then presses "Cancel" button, then opens the modal again - should the modal show previous state with deleted tags or not? Currently when tags are deleted but the change is not submitted/saved the modal does not show deleted tags.

src/pages/Tags/Tags.tsx
```diff
@@ -7,17 +7,11 @@ import DynamicForm, {
   FieldProps,
   InputTypes,
 } from 'components/organisms/DynamicForm/DynamicForm'
-// import { ReactComponent as Add } from 'assets/icons/add.svg'
+//import { ReactComponent as Add } from 'assets/icons/add.svg'
```
IngaSokman: I guess these outcommented rows are not needed

IngaSokman (changes requested):

src/components/molecules/EditableListContainer/EditableListContainer.tsx
```diff
@@ -0,0 +1,60 @@
...
+  data?: ListDataType[]
+  isEditable?: boolean
+  handelEditList?: () => void
+}
+
+const EditableListContainer: FC<EditableListTypes> = ({
+  title,
+  data,
+  isEditable,
+  handelEditList,
```
KaarelKa: handle* :D

src/components/molecules/TextInput/TextInput.tsx
```diff
@@ -67,7 +81,11 @@ const TextInput = forwardRef<
       <Label className={classNames(classes.label, !label && classes.hidden)}>
         {label}
       </Label>
-      <div className={classes.inputContainer}>
+      <div
+        className={classNames(classes.inputContainer, {
+          [classes.addDeleteButton]: true,
```
KaarelKa: **[non-blocker]**

If this class is always added, then I think you can just add it to the list, without adding the "true" condition

src/components/organisms/TagCategory/TagCategory.tsx
```diff
@@ -0,0 +1,89 @@
...
+        }
+      )
+
+      const payload: TagsUpdatePayloadType = {
+        type: values[0].type,
+        tags: updatedData,
+      }
+
+      try {
+        await updateTags(payload)
```
KaarelKa: **[non-blocker]**

It currently seems that we are taking the "type" argument from 2 different places. 1 comes from the passes in prop to TagCategory component and is used for part of the useBulkUpdate hook. the other one comes from the values added inside the modal. While this won't be an issue 99% of the time, I would still suggest using only 1 source of truth for the "type" value (in this case I would use the "type" passed in from props)

src/components/organisms/TagCategory/TagCategory.tsx
```diff
@@ -0,0 +1,89 @@
...
+        showNotification({
+          type: NotificationTypes.Success,
+          title: t('notification.announcement'),
+          content: t('success.tag_updated'),
+        })
+      } catch (errorData) {
+        showValidationErrorMessage(errorData)
```
KaarelKa: **[blocker]**

For forms we should show the error under the input, if we are dealing with validation errors and not close the modal by default. showValidationErrorMessage was meant for a few specific use cases where we had to show the alert instead (Seems that it is used it quite a few places, where it shouldn't be). Currently if the user tries to add a tag with the same name in the modal it is quite inconvenient. We close the modal and show them the alert, however this is just a validation error. Other validation errors (that are handled by FE) work differently (Modal stays open and error is displayed under the input)

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,236 @@
...
+  isLoading?: boolean
+  handelOnSubmit?: (values: EditDataType[]) => void
+  inputValidator?: (value?: string | undefined) => string | true
+}
+
+type EditableDefaultDataType = object & {
+  [key in string]?: string
+}
+
+type FormValues = EditableDefaultDataType
```
KaarelKa: **[blocker]**

Let's try to avoid these "Type A = Type B" assignments. Renaming a variable should serve some purpose, however right now it rather causes some confusion with EditableDefaultDataType and FormValues both being used, when actually the same type is meant

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,236 @@
...
+  closeModal: () => void
+  editableData?: EditDataType[]
+  title?: string
+  type?: TagTypes
+  isLoading?: boolean
+  handelOnSubmit?: (values: EditDataType[]) => void
+  inputValidator?: (value?: string | undefined) => string | true
+}
+
+type EditableDefaultDataType = object & {
```
KaarelKa: **[non-blocker]**

Not sure if this "object &" part is needed here, the type itself is essentially describing an object already (but might be wrong)

src/components/organisms/TagCategory/TagCategory.tsx
KaarelKa: **Extra note**

Once you add this logic, it means that we are not closing the modal, before we get the response from BE, so we should add some loading indicator (Most likely to the submit button)

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const defaultValues: EditableDefaultDataType = useMemo(
+    () =>
+      reduce(
+        editableData,
+        (result, value, key) => {
+          if (!value.id) {
+            return result
+          }
+          return {
+            ...(key > 0 && result),
```
KaarelKa: **[non-blocker]**

Not sure if this condition is necessary. I think just "...result" here would have the same effect. Might be missing something, but basically you are skipping the spreading of result, during the first iteration, when mapping over editableData. However result at that point would just be an empty object, so spreading it would result in nothing anyway.

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,236 @@
...
+        },
+        className: classes.editTagInput,
+        ...(includes(userPrivileges, Privileges.DeleteTag)
+          ? { handleDelete: () => handelOnDelete(`new_${size(inputFields)}`) }
+          : {}),
+      },
+    ])
+
+  const handelOnDelete = (name?: string, id?: string) => {
+    setDeletedValues((deletedValues) => [...deletedValues, { name, id }])
```
KaarelKa: **[non-blocker]**

Not a big issue, but would rename this "deletedValues" to "prevDeletedValues" or sth similar. Just to avoid any confusion between the actual "deletedValues" take from useState and the "deletedValues" here, which is basically the previous state. 

They both refer to the same data, but just for clarity it might be better

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  useEffect(() => {
+    const deleteFiled = filter(inputFields, (field) => {
```
KaarelKa: deletedFiles*

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const resetForm = useCallback(() => {
+    reset()
+    setDeletedValues([])
+  }, [editableFields, inputFields, reset])
+
+  const onSubmit: SubmitHandler<FormValues> = useCallback(
+    async (values) => {
+      const payload: EditDataType[] = map(values, (value, key) => {
+        switch (true) {
+          case !isEqual(values[key], defaultValues[key]) &&
```
KaarelKa: **[non-blocker]**

if you change the order of the first case and second case, you can skip the "!includes(key, 'new_')" check here

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,236 @@
...
+        }
+      })
+
+      try {
+        if (handelOnSubmit) {
+          handelOnSubmit(payload)
+        }
+      } catch (errorData) {
+        showValidationErrorMessage(errorData)
+      } finally {
```
KaarelKa: **[blocker]**

Mentioned previously about validation and closing of modal

src/hooks/requests/useTags.ts
```diff
@@ -67,3 +72,38 @@ export const useBulkCreate = () => {
...
+        (oldData?: TagsResponse) => {
+          const { data: previousData } = oldData || {}
+
+          if (!previousData) return oldData
+
+          const otherTypeData = filter(
+            previousData,
+            (data) => data.type !== type
+          )
+          const newData = [...otherTypeData, ...data]
```
KaarelKa, line 92: **[blocker]**

This is most likely the place, where the issue with order of TagContainers changing comes from (The one Inga mentioned with a screenshot). The tags will no longer be in the same order as they were before. This means that groupedCategoryData order will change as well.

This can be fixed either here, or possibly just by ordering groupedCategoryData (Probably easier + will resolve the case, when new TagCategories are added for the first time and the ordering issue with them)

src/pages/Tags/Tags.tsx
```diff
@@ -198,35 +189,21 @@ const Tags: FC = () => {
...
-                <li className={classes.tagName} key={tag?.id}>
-                  {tag?.name}
-                </li>
-              ))}
-            </ul>
-          </Container>
-        ))}
+        {map(
+          groupedCategoryData,
+          (tagsList: ListDataType[], type: TagTypes) => (
```
KaarelKa: **[blocker]**

Typing could be improved here. Currently you are rewriting tagsList to be of type ListDataType[] instead of the inferred TagFields[] type. Differences are that name and id are required in ListDataType and not required in TagFields. 

So if "id" and "name" are actually not needed in TagCategory, they shouldn't be required. If they are then we would still need to check for this, since we currently aren't checking for it and we know that in "tagsList" they are actually not required

If they should also be required in TagsList instead, then the change should be done there

src/types/tags.ts
```diff
@@ -5,12 +5,22 @@ export enum TagTypes {
   Oskused = 'Oskused',
 }
 
+export interface TagFields extends Object {
```
KaarelKa: **[blocker]**

This looks like a duplicate of the "Tag" type. What is the difference between them and should there be any ? They are both talking about the type of the same object

KaarelKa (changes requested):

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+  omitBy,
+  reduce,
+  size,
+  split,
+  toNumber,
+} from 'lodash'
+import { FieldPath, SubmitHandler, useForm } from 'react-hook-form'
+import { ReactComponent as Add } from 'assets/icons/add.svg'
+import { TagTypes } from 'types/tags'
+import useAuth from 'hooks/useAuth'
```
KaarelKa, line 33: **[non-blocker]** (For now)

You can deal with this in the departments task, where you plan to reuse this component, but currently you have the "TagTypes" type and Tag specific privilege checks inside the modal. Not sure if these will work with the department cases, so most likely you need to pass these as props as well.

But I haven't checked the spec for the department part, so not sure and either way it will be easier to deal with while dealing with departments

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+}
+
+export interface EditableListModalProps {
+  isModalOpen?: boolean
+  closeModal: () => void
+  editableData?: EditDataType[]
+  title?: string
+  type?: TagTypes
+  isLoading?: boolean
+  handleOnSubmit?: (values: EditDataType[]) => void
```
KaarelKa, line 59: **[non-blocker]**

I think some of these could be required props as well. For example the handleOnSubmit function should probably be a required prop

KaarelKa (approved):

PR #69, opened 2023-08-08
240

KaarelKa:


KaarelKa:
**Notes**

1. Task is semi-done, but most of the BE endpoints needed for this are either missing completely or not working correctly, so most of the stuff is mocked

2. Top section of new order view does not match the new design. It was completed before design changed. Will change it in the future if there is time

PR #71, opened 2023-08-11
180

VaabelMaret:


src/components/molecules/DepartmentManagement/DepartmentManagement.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {
+  useDepartmentsFetch,
+  useParallelMutationDepartment,
+} from 'hooks/requests/useDepartments'
+
+import EditableListContainer from 'components/molecules/EditableListContainer/EditableListContainer'
+import { ModalTypes, showModal } from 'components/organisms/modals/ModalRoot'
+import { EditDataType } from 'components/organisms/modals/EditableListModal/EditableListModal'
+import { showNotification } from 'components/organisms/NotificationRoot/NotificationRoot'
+import { NotificationTypes } from '../Notification/Notification'
```
KaarelKa: **[non-blocker]**

Not too important, but would be good to use absolute imports everywhere (it auto-import with relative import if you are in the same subfolder under components).

Won't change much for now, but we might start reorganizing the files at some point and it's a lot easier to do with absolute imports

src/components/molecules/DepartmentManagement/DepartmentManagement.tsx
```diff
@@ -0,0 +1,60 @@
...
+    },
+    [parallelUpdating, t]
+  )
+
+  const handleEditDepartmentsModal = () => {
+    showModal(ModalTypes.EditableListModal, {
+      editableData: existingDepartments,
+      title: t('modal.edit_departments'),
+      handleOnSubmit: handleOnSubmit,
+      isLoading,
```
KaarelKa: **[blocker]**

Didn't test, but I have a feeling that this "isLoading" value doesn't work. It will be "false" when the button is clicked and modal is opened. Even if it turns to "true" while the modal is open, the modal won't get that update, unless it is opened again with handleEditDepartmentsModal

src/components/molecules/DepartmentManagement/DepartmentManagement.tsx
```diff
@@ -0,0 +1,60 @@
...
+        Privileges.DeleteDepartment
+      ),
+    })
+  }
+
+  return (
+    <EditableListContainer
+      title={t('cheat_sheet.institution_management.departments')}
+      data={existingDepartments}
+      isEditable={true}
```
KaarelKa: **[non-blocker]**

Possibility to just use "isEditable" shorthand here

KaarelKa (changes requested):

KaarelKa:
has some merge conflicts

src/components/organisms/modals/EditableListModal/EditableListModal.tsx
```diff
@@ -257,7 +294,7 @@ const EditableListModal: FC<EditableListModalProps> = ({
         icon={Add}
         children={t('tag.add_new_row')}
         onClick={addInputField}
```
KaarelKa, line 318: **[blocker]**

There seems to be an issue with addInputField. Since it uses the length the the array for input names, we will start getting duplicated names, if we delete some random rows from this modal and try to add new ones after that.

src/components/molecules/DepartmentManagement/DepartmentManagement.tsx
```diff
@@ -0,0 +1,60 @@
...
+    },
+    [parallelUpdating, t]
+  )
+
+  const handleEditDepartmentsModal = () => {
+    showModal(ModalTypes.EditableListModal, {
+      editableData: existingDepartments,
+      title: t('modal.edit_departments'),
+      handleOnSubmit: handleOnSubmit,
+      isLoading: isLoading,
```
KaarelKa: **[blocker]**

isLoading issue which was mentioned previously is still here. You are passing "isLoading" false to the modal, when you open it and this prop will never get updated inside the modal, so modal will never have "isLoading: true".

src/components/organisms/TagCategory/TagCategory.tsx
```diff
@@ -63,6 +66,8 @@ const TagCategory: FC<TagCategoryTypes> = ({ tagsList, type, isEditable }) => {
       type,
       isLoading: isUpdatingTags,
```
KaarelKa: **[blocker]**

Seems that we have the same isLoading issue here as well

src/components/molecules/DepartmentManagement/DepartmentManagement.tsx
```diff
@@ -0,0 +1,60 @@
+import { FC, useCallback } from 'react'
```
KaarelKa, line 1: **[blocker]** There seem to be some other issues with changing :

1. When I try to add 1 new department and I leave it empty, I will get a success notification, the modal will close, but the new department with empty name is not added. If I'm not mistaken we should show error there instead and not allow submitting the form.

2. Similar issue, when I change the name of an existing department to an empty string. It will show a success notification, as if everything worked, however that department is actually just ignored and the value doesn't get changed + user won't see an error

KaarelKa (changes requested):

src/components/molecules/DepartmentManagement/DepartmentManagement.tsx
KaarelKa: **[blocker]**

This issue still exists to some degree, it allows be to submit a form with new empty departments and changing existing departments to empty strings, however it does nothing other than closing the modal. This can be very confusing for a user (No idea if the changes went through or not + once they realise the changes didn't go through they still don't know why)

src/hooks/requests/useDepartments.ts
```diff
@@ -22,3 +24,73 @@ export const useDepartmentsFetch = () => {
...
+      reject([...errors, { values: fulfilled }])
+    }
+  })
+
+export const useParallelMutationDepartment = () => {
+  const queryClient = useQueryClient()
+  const { mutateAsync: parallelUpdating, isLoading } = useMutation({
+    mutationKey: ['departments'],
+    mutationFn: async (payload: EditDataType[]) =>
+      requestsPromiseThatThrowsAnErrorWhenSomeRequestsFailed(payload),
```
KaarelKa, line 86: **[non-blocker]** (Would be a blocker actually)

Basically the ordering of these requests should be revised. At the moment If I try to add a tag with name "Veel 2", but I also remove the previous "Veel 2" department from the list, the deletion will go through, but adding won't since the adding part happens before deletion in some cases, which means that at that point BE thinks that we already have the "Veel 2" tag.

But let's leave it for now, there is no point in putting more time under this part, since we will change the endpoint we use, once we have the bulk edit endpoint anyway.

KaarelKa (changes requested):

src/components/molecules/DepartmentManagement/DepartmentManagement.tsx
VaabelMaret: I'm currently adding a "No Changes Made" notification. I think if we can find a solution to use react-hook-form useFieldArray to add and remove values and fields, then we can use submitt button disable solution

VaabelMaret:
Fixed privileges bug task 446 here also.
https://github.com/keeleinstituut/tv-tolkevarav/issues/446

PR #75, opened 2023-08-17
157

IngaSokman:


src/components/atoms/LanguageLabels/LanguageLabels.tsx
```diff
@@ -0,0 +1,56 @@
+import { FC } from 'react'
+import { useTranslation } from 'react-i18next'
+import { find, map } from 'lodash'
+import { useWatch } from 'react-hook-form'
+import { VendorPriceListSecondStepProps } from 'components/organisms/forms/VendorPriceListForm/VendorPriceListForm'
+
+import classes from './classes.module.scss'
+
+type LanguageLabelsProps = Omit<
```
KaarelKa, line 9: **[non-blocker]**

It would be better to have parents extend the prop types of their children rather than children extending the prop types of their parent. Currently you have VendorPriceListThirdStepProps and VendorPriceListSecondStepProps, which both use LanguageLabels as a child and accept all the props that LanguageLabels accepts + VendorPriceListSecondStepProps accepts 1 additional prop.

So if we would have parent extending props of children, we would just have VendorPriceListSecondStepProps and VendorPriceListThirdStepProps extending the props of LanguageLabels + no need to omit anything. 

In the current case it doesn't have any direct downsides, since the extra prop is Omitted, however for consistency + for cases, where we might want to use LanguageLabels in some other component, it would still be good to keep the same direction of extending types everywhere

src/components/organisms/forms/VendorPriceListForm/VendorPriceListForm.tsx
```diff
@@ -0,0 +1,442 @@
...
+      try {
+        await createPrices(payload)
+
+        showNotification({
+          type: NotificationTypes.Success,
+          title: t('notification.announcement'),
+          content: t('success.language_pairs_prices_added'),
+        })
+        resetForm()
+      } catch (errorData) {
```
KaarelKa, line 346: **[blocker}**

If we get validation errors we should be showing them under the corresponding fields inside the modal and we shouldn't close the modal. Currently if the user makes a mistake, they have to start the entire flow again, but they should be allowed to fix the mistakes they made. This also means that if the first error was on step 1, then we should show them the first step, instead of the last.

Considering this, we also have https://github.com/keeleinstituut/tv-tolkevarav/issues/426 task for navigating between the steps, so if you want, you can do the 426 task within this task, but it's not necessary.

src/components/organisms/tables/AddVendorPricesTable/AddVendorPricesTable.tsx
```diff
@@ -0,0 +1,189 @@
...
+            ariaLabel={t('vendors.minimal_fee')}
+            type="number"
+            className={classes.pricesInput}
+            placeholder="00.00€"
+          />
+        )
+      },
+      footer: (info) => info.column.id,
+    }),
+  ] as ColumnDef<any>[]
```
KaarelKa, line 170: **[non-blocker]**

Do we need this any type here ?

src/types/vendors.ts
```diff
@@ -53,3 +53,107 @@ export type UpdateVendorPayload = {
...
+  dst_lang_classifier_value_id: string
+  updated_at: string
+  created_at: string
+  character_fee: number
+  word_fee: number
+  page_fee: number
+  minute_fee: number
+  hour_fee: number
+  minimal_fee: number
+  vendor: {
```
KaarelKa, line 97: **[blocker]** Let's use existing types if they already exists. We have the "Vendor" type defined under types/vendors

KaarelKa (changes requested):

KaarelKa:
It seems we also have a couple of merge conflicts

PR #81, opened 2023-08-24
Fix-449-added new title to tag modals and removed skills from options

VaabelMaret:


KaarelKa:
@VaabelMaret This can be merged after merge conflict is fixed

PR #82, opened 2023-08-24
235

IngaSokman:


IngaSokman:
Currenlty added pagination even though it wasn't in the design.

IngaSokman:
Ready for review

KaarelKa:
**[blocker]**

Currently can't send the form when trying to add something. (Note: couldn't reproduce every time, I think it will happen, when you play around with the skills a bit - tick a few, untick some of them, etc)

1. It seems that every CheckBox field under skills is required, so the only way to save some prices is to pick every skill.

2. I can also see the error that should be attached to the Checkbox inputs (The isRequired error), currently shows up somewhere else (sometime behind the modal, sometimes ontop of the modal, but positioned nearly at the top left of the screen). -- **[note]** No need to deal with this actually, I'll fix it at some point, it's not related to this task

KaarelKa:
**[blocker]**

When adding prices, it seems that I can't save the form again, if I get some validation errors the first time, even after fixing them.

Repro:

1. Filled in the form, but used the same source and dst language
2. Submitted the form and got an error
3. Got navigated back to the first step and error is displayed (all good so far)
4. Change the destination language
5. Navigated back to the last step of form and pressed submit button
6. Nothing happens

KaarelKa:
**[blocker]**

1. Error not displayed under the input, when editing some prices
2. I can enter random letters into the price field (at least when editing). -- (Can be fixed later)

KaarelKa:
**[blocker]**

Some weird behaviour is happening, when I edit 1 of the prices or delete 1 of them. Previously it showed be 8 language pairs on the first page (I only added 8), but after the edit/delete it only shows be 5 on 1 page and the rest moved to the next page

KaarelKa:
**[blocker]**

When adding prices there seems to be 1 error case, which is not handled. If you try to add prices for the same skill in the same language pair, BE gives an error ("Price already exists"). We should also show those. Makes sense to show this error under the checkboxes as well in step 2

KaarelKa:
**[blocker]**

It seems that when adding prices, some of the option in the prices table are missing. Picked 12 skills, but I only see 10 rows for setting prices.

src/components/atoms/LanguageLabels/LanguageLabels.tsx
```diff
@@ -0,0 +1,79 @@
...
+          }
+        >
+          {srcLanguageLabel}
+        </p>
+      </div>
+      <div className={classes.languageContianer}>
+        <p className={classes.destinationLanguage}>
+          {`${t('vendors.destination_language')}*`}
+        </p>
+        {map(dstLanguageLabels, (label, index) => (
```
KaarelKa, line 72: **[non-blocker]**

I think our "Tag" component could have been used here as well, but not too important at this stage

src/components/atoms/ProgressBar/ProgressBar.tsx
```diff
@@ -9,9 +9,18 @@ export interface Step {
 interface ProgressBarProps {
   steps: Step[]
   activeStep: number
+  setActiveStep: React.Dispatch<React.SetStateAction<number>>
```
KaarelKa: **[non-blocker]**

Let's try to avoid using these heavily restricted types, just in case we need to use this component anywhere else. In most other places where these restrictions were used, we already had to change them. The way setActiveStep is used, is just as function, which accepts a number as an argument and return nothing.

src/components/atoms/ProgressBar/ProgressBar.tsx
```diff
@@ -9,9 +9,18 @@ export interface Step {
...
 }
 
-const ProgressBar: FC<ProgressBarProps> = ({ steps, activeStep }) => {
+const ProgressBar: FC<ProgressBarProps> = ({
+  steps,
+  activeStep,
+  setActiveStep,
+}) => {
+  const handleStepClick = (index: number) => {
```
KaarelKa: **[non-blocker]**

Seems like an unnecessary middle-man function. Its basically just a new function that does nothing except calling another function with exactly the same argument

src/components/organisms/AddVendorPriceModalButton/AddVendorPriceModalButton.tsx
```diff
@@ -0,0 +1,271 @@
...
+            />
+          ),
+        },
+      ],
+      submitForm: handleSubmit(onAddPricesSubmit),
+      resetForm: resetForm(),
+      buttonComponent: (
+        <VendorPriceListButtons
+          control={control}
+          isLoading={isCreatingPrices}
```
KaarelKa: **[blocker]**

isLoading won't work like this, you can also discuss this with @VaabelMaret , we recently went over the same thing. isCreatingPrices will have some value (mostly false), when "handleAddPriceModal" function is executed (Button is pressed).

The button "VendorPriceListButtons" will be inside the modal, with "isLoading" value fixed to whatever it was, when the modal opened. Even though the value of isCreatingPrices might change later or, you are not sending the updated value to the Modal, or rather the instance of VendorPriceListButtons inside the modal, which means it will never change.

You can read more about this here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures

src/components/organisms/AddVendorPriceModalButton/AddVendorPriceModalButton.tsx
```diff
@@ -0,0 +1,271 @@
...
+  const { createPrices, isLoading: isCreatingPrices } =
+    useCreatePrices(vendorId)
+
+  const { classifierValuesFilters: languageFilter } = useClassifierValuesFetch({
+    type: ClassifierValueType.Language,
+  })
+
+  const { data: skillsData } = useFetchSkills()
+
+  const languageOptions = map(languageFilter, ({ value, label }) => {
```
KaarelKa: **[blocker]**

This mapping here is a bit confusing. You are mapping over an array of object, which have "value" and "label" and are returning an array of identical objects. Seems very unnecessary

src/components/organisms/AddVendorPriceModalButton/AddVendorPriceModalButton.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+  const skillsFormFields: FieldProps<FormValues>[] = map(
+    skillsData,
+    ({ id, name }, index) => {
+      return {
+        key: index,
+        inputType: InputTypes.Checkbox,
+        ariaLabel: name || '',
+        label: name,
+        name: `skill_id.${id}_${index}`,
```
KaarelKa: **[non-blocker]**

Not sure if we need the index part here. "id" should already be unique + currently I just see that the index is some extra part of the name string that needs to be removed later (when submitting the form). But it can stay as well, if it bring any extra value/makes something simpler.

src/components/organisms/DeleteVendorPriceButton/DeleteVendorPriceButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+        type: NotificationTypes.Success,
+        title: t('notification.announcement'),
+        content: t('success.language_pairs_prices_deleted'),
+      })
+    } catch (errorData) {
+      showValidationErrorMessage(errorData)
+    }
+  }, [deletePrices, languagePairModalContent, t])
+
+  const handleDelete = () => {
```
KaarelKa: **[blocker]**

Seems like an unnecessary function. It just calls another function and does nothing else

src/components/organisms/DeleteVendorPriceButton/DeleteVendorPriceButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+    onDeletePrices({})
+  }
+
+  return (
+    <Button
+      appearance={AppearanceTypes.Text}
+      icon={Delete}
+      ariaLabel={t('vendors.delete')}
+      onClick={() => handleDelete()}
```
KaarelKa: **[blocker]**

No need to create a new anonymous function for calling a different function, especially when we are not passing any arguments.

Basically here we could just say that "onClick" should use the "onDeletePrices" function.

Instead, when the user presses the button, we.

1. First create a new anonymous function and call handleDelete in that function
2. Inside handleDelete we call the onDeletePrices function
3. Inside onDeletePrices we do what we actually wanted to do

src/components/organisms/DeleteVendorPriceButton/DeleteVendorPriceButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+  languagePairModalContent,
+  vendorId,
+}) => {
+  const { t } = useTranslation()
+  const { userPrivileges } = useAuth()
+
+  const { deletePrices, isLoading: isDeletingPrices } =
+    useDeletePrices(vendorId)
+
+  const onDeletePrices: SubmitHandler<FormValues> = useCallback(async () => {
```
KaarelKa: **[blocker]** 

Typing of this onDeletePrices function seems to be wrong. It doesn't look like it's being used to submit any form. This component doesn't seem to be related to the VendorPriceListForm at all actually

src/components/organisms/EditVendorPriceModalButton/EditVendorPriceModalButton.tsx
```diff
@@ -0,0 +1,177 @@
...
+      helperText: t('vendors.price_list_change_description'),
+      modalContent: (
+        <VendorPriceListEditContent
+          control={control}
+          editableSkills={languagePairModalData}
+          srcLanguageValue={srcLanguageValue}
+          dstLanguageValues={[dstLanguageValue]}
+        />
+      ),
+      isLoading: isUpdatingPrices,
```
KaarelKa: **[blocker]**

Same issue with isLoading here

src/components/organisms/EditVendorPriceModalButton/EditVendorPriceModalButton.tsx
```diff
@@ -0,0 +1,177 @@
...
+        character_fee,
+        hour_fee,
+        minute_fee,
+        minimal_fee,
+        page_fee,
+        word_fee,
+        skill,
+      }) => {
+        return {
+          id: id,
```
KaarelKa: **[non-blocker]**

Would suggest using shorthand for cases where the variable name is the same as the key you want to use for it + if there is no need to change the value inside an object.

`
{
    id,
    character_fee,
    ...
}
`

instead of:


`
{
    id: id,
    character_fee: character_fee,
    ...
}
`

src/components/organisms/forms/VendorPriceListForm/VendorPriceListForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+    resetForm()
+  }, [resetForm])
+
+  const columnHelper = createColumnHelper<PriceObject>()
+
+  const columns = [
+    columnHelper.accessor('language_direction', {
+      header: () => t('vendors.language_direction'),
+      cell: ({ row }) => {
+        // eslint-disable-next-line react-hooks/rules-of-hooks
```
KaarelKa: **[blocker]**

This can cause some annoying errors with the way react works internally. Can't really predict them right now, but would feel a lot better if we could move this useEffect into some correct place. For example by creating a small extra component in this file for this specific cell, passing row to it and using useEffect in there

src/components/organisms/forms/VendorPriceListForm/VendorPriceListForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+            />
+            <DeleteVendorPriceButton
+              languagePairModalContent={languagePairModalContent}
+              vendorId={vendor_id}
+            />
+          </div>
+        )
+      },
+    }),
+  ] as ColumnDef<any>[]
```
KaarelKa, line 319: **[non-blocker]**

Would be good to type this as well, but if it looks like too much of an hassle, add some ignore line here, so tslint would ignore this warning.

src/components/organisms/forms/VendorPriceListForm/VendorPriceListForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+          className={
+            !isEmpty(groupedLanguagePairData)
+              ? classes.vendorPricesContainer
+              : classes.hiddenVendorPrices
+          }
+          paginationData={paginationData}
+          onPaginationChange={handlePaginationChange}
+          title={
+            <div className={classes.pricesDataTableHeader}>
+              {t('vendors.vendor_price_list_title')}
```
KaarelKa: **[blocker]**

Text should always be in appropriate html tags

src/components/organisms/modals/FormProgressModal/FormProgressModal.tsx
```diff
@@ -1,80 +1,114 @@
...
   TitleFontTypes,
 } from 'components/organisms/ModalBase/ModalBase'
-import { AppearanceTypes } from 'components/molecules/Button/Button'
-import { useTranslation } from 'react-i18next'
-import { FC, ReactElement, useState } from 'react'
+import { FC, ReactElement, useEffect, useState } from 'react'
 import ProgressBar from 'components/atoms/ProgressBar/ProgressBar'
-import { find, map, size } from 'lodash'
+import { find, keys, map, size } from 'lodash'
+import { FormValues } from 'components/organisms/forms/VendorPriceListForm/VendorPriceListForm'
```
KaarelKa, line 9: **[non-blocker]**

FormProgressModal was actually meant for any progress modal we might have, but currently it won't work for any other use case, since it will only work with the "FormValues" form values type from VendorPriceListForm. 

It should instead accept FormValues as a type argument.

Currently this modal is a bit confusing. It's now built in a way that it can't really be reused for any other situation, but it still accepts some things as option props, even though it is only used for 1 specific use case, where we already know all these values. For example buttonComponent prop seems a bit unnecessary, since we will only use this modal for adding Vendors price list and we already know what the button should be at what it should be.

Normally we have 1 abstract component for reusing (For example ModalBase) and then some components that utilize it for specific use cases (For example TooltipModal). Main benefits that come from those are that abstract components usually try to be stupid components that contain as much repetitive/shared code as possible and the components made for specific use cases should contain all the logic they need for that case inside themselves, so they can be easily reused in every place, where we need that logic.

Currently this component looks like both at the same time, which in the end will mean that it doesn't really get the benefits of either. It can't be reused for anything other than adding prices to a vendor, but at the same time it needs a lot of props passed in (or rather some props that need to contain a lot of information, like the entire formData), so it can't be easily reused if we would want to add some prices to vendors from some other place in the project.

src/components/organisms/modals/FormProgressModal/FormProgressModal.tsx
```diff
@@ -1,80 +1,114 @@
...
+import React from 'react'
```
KaarelKa: **[non-blocker]**

1. Would be good to keep all the imports from the same library on 1 row (React in this case).
2. For consistency it would make more sense to import "cloneElement" as a named import from react the same way we import everything else from react

src/components/organisms/modals/FormProgressModal/FormProgressModal.tsx
```diff
@@ -1,80 +1,114 @@
...
+    : null
+
   return (
     <ModalBase
       title={title}
       titleFont={TitleFontTypes.Gray}
       open={!!isModalOpen}
       buttonsPosition={ButtonPositionTypes.SpaceBetween}
       progressBar={
-        steps && <ProgressBar activeStep={activeStep} steps={steps} />
```
KaarelKa, line 62: **[non-blocker]**

No need to check for the existence of "steps" here, steps is constructed in this file and will always be truthy

src/components/organisms/modals/FormProgressModal/FormProgressModal.tsx
KaarelKa: No need to do changes at this point, but for future components, I would suggest keeping this in the back of your mind and always deciding what type of component you are currently building. Sometimes during development they get mixed up a bit, but then they should be cleaned up afterwards and separated. 

At this stage in the project we can leave them, since we don't have a lot of time, but just wanted to mention this for future projects/components

src/components/organisms/tables/AddVendorPricesTable/AddVendorPricesTable.tsx
```diff
@@ -0,0 +1,195 @@
...
+      footer: (info) => info.column.id,
+    }),
+    columnHelper.accessor('character_fee', {
+      header: () => t('vendors.character_fee'),
+      cell: ({ row }) => {
+        return (
+          <FormInput
+            key={row?.index}
+            name={`character_fee-${row?.index}`}
+            control={control}
```
KaarelKa: **[non-blocker]**

Would be easier to use, if we added some masking to these inputs, but this is nice to have at this point

src/hooks/requests/useVendors.ts
```diff
@@ -62,3 +70,125 @@ export const useUpdateVendor = (vendorId: string) => {
     isLoading,
   }
 }
+
+export const useFetchSkills = () => {
+  const { isLoading, isError, data } = useQuery<GetSkillsPayload>({
```
KaarelKa: **[non-blocker]**

Not too important, but the type argument passed to useQuery should actually be the type of the response we get from BE. so basically the content of GetSkillsPayload type is correct, but the naming could use "Response" instead of "Payload"

src/types/vendors.ts
```diff
@@ -53,3 +53,101 @@ export type UpdateVendorPayload = {
...
+export type SkillsData = {
+  id: string
+  name: string
+}
+
+export type GetSkillsPayload = {
+  data?: SkillsData[]
+}
+
+export interface Prices {
```
KaarelKa: **[blocker]** (potentially)

Looks essentially like a duplicate of the "Price" type we have under "types/price", although I see some small differences. But pretty sure these should be the same type

src/types/vendors.ts
```diff
@@ -53,3 +53,101 @@ export type UpdateVendorPayload = {
...
+}
+
+export type UpdatePricesPayload = {
+  data?: UpdatedPricesData[]
+}
+export type DeletePricesPayload = {
+  id?: string[]
+}
+
+export type PricesData = {
```
KaarelKa: **[blocker]**

Hmm, this looks a lot like another copy of the "Prices" type here and the "Price" type in "types/price".

These should be unified. If the difference is that from some endpoint we get less fields and from other endpoints we get more fields then we should have 1 type defined with the fields we always get and 1 extra type that extends the first one, with the new fields added.

KaarelKa (changes requested):

src/components/organisms/modals/FormProgressModal/FormProgressModal.tsx
```diff
@@ -1,80 +1,113 @@
...
-
   const handleProceed = () => {
     if (size(steps) === activeStep) {
-      //Todo add form submit
-      // submitForm()
-      closeModal()
-      setActiveStep(1)
+      if (submitForm) {
+        submitForm()
+        setActiveStep(1)
```
KaarelKa: **[blocker]**

Currently this will navigate the user to the first step, even if submitting the form was successful. In case of success the form will close briefly afterwards, so it looks like a flickering bug.

src/components/organisms/modals/FormProgressModal/FormProgressModal.tsx
```diff
@@ -1,80 +1,113 @@
...
+  control,
 }) => {
-  const { t } = useTranslation()
   const [activeStep, setActiveStep] = useState(1)
+
+  const formState = useFormState({ control })
+  const formErrors = formState.errors
+  const valuesKey = keys(formErrors)[0]
+
+  const isErrorOnSecondStep =
```
KaarelKa: **[blocker]**

Currently it takes me to second step, when I try to add another skill to an existing language pair (Add not edit). However it looks like it should take me to the first step instead, since the actual issue is with the language pair.

KaarelKa:
**[blocker]** had the issue with submitting after fixing again, kind of. Currently the issue was that I tried to add 1 skill to a language pair that already existed. After getting the error, I changed the destination language, so the form state should actually be valid. If I now press the submit button in the last step, it just navigates me back to the first step. It seems that the other fields will still remain in an error state, since I didn't touch them.

src/components/atoms/LanguageLabels/LanguageLabels.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  const sourceLanguageLabel = findLabelByValue(
+    useWatch({ control, name: 'src_lang_classifier_value_id' })
+  )
+  const destinationLanguageLabels = findLabelByValue(
+    useWatch({ control, name: 'dst_lang_classifier_value_id' })
+  )
+
+  const srcLanguageLabel = srcLanguageValue
```
KaarelKa: **[non-blocker]** Could also use srcLanguageValue || sourceLanguageLabel here, which would be slightly less code

src/components/organisms/DeleteVendorPriceButton/DeleteVendorPriceButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+      showValidationErrorMessage(errorData)
+    }
+  }, [deletePrices, languagePairModalContent, t])
+
+  return (
+    <Button
+      appearance={AppearanceTypes.Text}
+      icon={Delete}
+      ariaLabel={t('vendors.delete')}
+      onClick={() => onDeletePrices()}
```
KaarelKa: **[non-blocker]**

No need to create a new anonymous function that just calls onDeletePrices function here

src/components/organisms/EditVendorPriceModalButton/EditVendorPriceModalButton.tsx
```diff
@@ -0,0 +1,181 @@
...
+    })
+  }
+
+  return (
+    <Button
+      appearance={AppearanceTypes.Text}
+      icon={Edit}
+      ariaLabel={t('vendors.edit_language_pair')}
+      className={classes.editIcon}
+      onClick={() => handleEditPriceModal(languagePairModalContent)}
```
KaarelKa: **[blocker]**

It seems that currently editing of skills attached to a language pair is not added. There was discussion about it here: https://www.figma.com/file/VqCwED3Dymn7L1gGoAOK27?node-id=919:62333&mode=dev#407299699

src/types/vendors.ts
```diff
@@ -89,3 +89,103 @@ export type UpdateVendorPayload = {
...
+  id?: string
+  character_fee?: number
+  word_fee?: number
+  page_fee?: number
+  minute_fee?: number
+  hour_fee?: number
+  minimal_fee?: number
+}
+
+export type Price2 = {
```
KaarelKa: **[blocker]**

Why do we need duplicated types for Price ?

src/components/organisms/DeleteVendorPriceButton/DeleteVendorPriceButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+}) => {
+  const { t } = useTranslation()
+  const { userPrivileges } = useAuth()
+
+  const { deletePrices, isLoading: isDeletingPrices } =
+    useDeletePrices(vendorId)
+
+  const onDeletePrices = useCallback(async () => {
+    const payload = {
+      id: map(languagePairModalContent, ({ id }) => id),
```
KaarelKa: **[blocker]**

I think this will always delete all the skills under 1 language pair, but it should depend on which delete button is pressed. If the delete button next to the language pair is clicked, then it should delete all of them, but if the delete button on 1 of the skill rows is pressed, it should only remove that skill.

KaarelKa (changes requested):

IngaSokman:
Ready for review, but note,  error message from BE regarding inputfield does not show it currenlty. There is a task for it - [490](https://github.com/orgs/keeleinstituut/projects/2/views/3?filterQuery=no%3Aepic+-label%3Astory+super&pane=issue&itemId=41118150)

IngaSokman:
NB! Currently error message coming from validation is shown a a bit lower than it should be, but this should be fixed on task 347

superspartak (approved):

src/api/ApiClient.ts
```diff
@@ -49,10 +49,10 @@ class ApiClient {
   debug = (title: string, content: object) => {
     if (this.debug_mode && (title || content)) {
       if (title) {
-        console.log(`=== DEBUG: ${title} ===========================`)
+        // console.log(`=== DEBUG: ${title} ===========================`)
```
superspartak, line 52: Are they still needed?

src/components/organisms/ModalBase/ModalBase.tsx
```diff
@@ -67,21 +68,26 @@ export interface ModalProps extends ModalFooterProps {
...
+    <>
+      <div className={classes[buttonsPosition]}>
+        {buttons &&
+          map(buttons, (button, index) => (
+            <Button key={index} {...button}>
+              {button?.children}
+            </Button>
+          ))}
+      </div>
+      {buttonComponent && buttonComponent}
```
superspartak: What does it mean?

src/api/ApiClient.ts
IngaSokman: For debugging purposes it is nice to have

src/components/organisms/ModalBase/ModalBase.tsx
IngaSokman: This is added from FormProgressModal for the buttons to work properly

PR #85, opened 2023-08-31
Manual

IngaSokman:


IngaSokman:
Ready for review

src/components/organisms/Tooltip/Tooltip.tsx
```diff
@@ -1,20 +1,29 @@
...
+import helpSections from 'pages/Manual/helpSections.json'
 
 import classes from './classes.module.scss'
-import { useTranslation } from 'react-i18next'
+
+export interface HelpSections {
+  [key: string]: {
+    title: string
+    content: string
+    tooltipContent: string
```
KaarelKa: **[non-blocker]**

Let's have "tooltipContent" as an optional field. We might have sections in the manual that won't have a matching tooltip, in which case they also might not have "tooltipContent"

src/components/organisms/modals/TooltipModal/classes.module.scss
```diff
@@ -10,6 +10,10 @@
 .contentStyle {
```
KaarelKa, line 11: **[blocker]**

Seems that the list ( ul/li tags ) styling we currently have here doesn't match the design.

1. One thing we should do here is create some mixin (under styles/mixins) for the styling we want to apply to these helper html sections. This will not use any classNames, since the content we get will be pure html, so it will just have styles for html tags. This way we can include the exact same styles for both the manual page and this modal.

2. Once the styles in 1. are created, they can be included under the .contentStyle class in the TooltipModal.

3. On the manual page we should also wrap the {manualContent} with some div and give a className to that div, so we can add the same styles there with the mixin.

4. For now you can just check some of the example we have in design for the styles that should be applied. Mostly everything else seems fine, just the ul/li part might need some work. Later there might be new tags used in the manual page and we might need to edit those styles (This is why it would be good to have a mixin for it that both the tooltip and manual page use)

Screenshot 2023-09-06 at 10 09 06

src/components/templates/AuthWrapper/AuthWrapper.tsx
```diff
@@ -22,12 +22,14 @@ const AuthWrapper: FC<PropsWithChildren> = () => {
         institutionUserId: userInfo?.tolkevarav?.institutionUserId || '',
       }}
     >
-      <Loader
+      {/* <Loader
```
KaarelKa: **[blocker]**

Let's add these back before merging. This is fine for local development, but all temporary changes used to allow us to develop should be removed before merging to dev, since they will be hard to spot later on

src/pages/Manual/Manual.tsx
```diff
@@ -1,9 +1,90 @@
...
+          <p className={classes.articleInfo}>
+            {t('menu.article_created_by', { author, authorEmail })}
+          </p>
+          <p className={classes.articleInfo}>
+            {t('menu.last_changed', { date })}
+          </p>
+        </div>
+        {manualContent}
+      </div>
+      <div className={classes.manualTitleContainer}>
```
KaarelKa, line 82: **[blocker]**

The table of contents should always be visible and should be separately scrollable (Similar to how our SideBar works).

KaarelKa:
Added a few comments, but generally good job, this is pretty much exactly what I had in mind

KaarelKa (changes requested):

IngaSokman:
Ready for review

src/i18n/locales/et.json
```diff
@@ -188,7 +188,10 @@
     "institution_settings": "Asutuse sätted",
     "technical_settings": "Tehnilised sätted",
     "collapse_menu": "Kitsendan menüü",
-    "manual": "Kasutusjuhend"
+    "manual": "Tõlkevärava kasutusjuhend",
```
KaarelKa, line 191: **[blocker]** 

We should add a new string for this, or a new one for the sidebar. In sidebar the item should be called "Kasutusjuhend"

src/pages/Manual/Manual.tsx
KaarelKa: **[blocker]**

It currently seems that part of the content in the table of content will be hidden. Added some extra items and currently the bottom 2 will be out of screen and can't be scrolled to.

KaarelKa (changes requested):

src/pages/Manual/Manual.tsx
IngaSokman: Missed that, fixed

PR #89, opened 2023-09-14
213

VaabelMaret:


src/hooks/useValidators.ts
```diff
@@ -72,13 +72,26 @@ const useValidators = () => {
     return true
   }
 
+  const dateTimeValidator = (value?: any) => {
+    if (
+      !value.days ||
```
superspartak: it will throw an error if value is undefined. Maybe `!value?.days` ?

src/pages/InstitutionSettings/InstitutionSettings.tsx
```diff
@@ -30,7 +31,15 @@ const InstitutionSettings: FC = () => {
...
         <h3 className={classes.title}>{t('institution.institution_data')}</h3>
-        <InstitutionForm name={name} id={institutionId} {...institution} />
+        <InstitutionForm
+          name={name}
+          id={institutionId}
+          {...institution}
+          workingTimes={
+            <WorkingTimes name={name} id={institutionId} data={institution} />
+          }
+          //vacationDays={<div>Puhkepäevad:</div>}
```
superspartak, line 42: Commented out, left in

src/components/organisms/DynamicInputComponent/DynamicInputComponent.tsx
```diff
@@ -52,6 +61,9 @@ export enum InputTypes {
   Time = 'time',
   TagsSelect = 'tagsSelect',
   DateTime = 'dateTime',
+  TimeRange = 'timeRange',
+  DayTimeRange = 'DayTimeRange',
```
superspartak: For sake of consistency, don't capitalise the values of enums

superspartak:
In range pickers, does it make sense to implement some restriction so that start time/date cannot be after the end time/date?

src/pages/InstitutionSettings/InstitutionSettings.tsx
VaabelMaret: This is needed. It's TODO, I will add TODO comment as well

VaabelMaret:
Currently, BE sends an error message if the customer has selected an incorrect range. I don't know if it's needed in FE side as well. On the FE side, I show the errors messages and the customer can fix them and submit the correct ones again.

src/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,103 @@
+import DatePickerInput from 'components/molecules/DatePickerInput/DatePickerInput'
+import { Ref, forwardRef, useCallback } from 'react'
+import { FieldError } from 'react-hook-form'
+import classNames from 'classnames'
+import { ReactComponent as Delete } from 'assets/icons/delete.svg'
+import classes from './classes.module.scss'
+import { useTranslation } from 'react-i18next'
+import Button, { AppearanceTypes, IconPositioningTypes } from '../Button/Button'
```
IngaSokman, line 8: **[non-blocker]** Not too important but we should use absolute imports everywhere

IngaSokman (approved):

PR #90, opened 2023-09-15
161

IngaSokman:


src/pages/GeneralPriceList/GeneralPriceList.tsx
```diff
@@ -0,0 +1,87 @@
...
+            className={classes.searchInput}
+            inputContainerClassName={classes.generalPriceListInput}
+            isSearch
+          />
+        </Root>
+      </div>
+      <Loader loading={isLoading} />
+      <GeneralPriceListTable
+        data={prices}
+        hidden={isLoading}
```
VaabelMaret: [blocker]

the table jumps and flashes when using sorting and filtering. It's better to not use hidden prop and add column size to avoid the jumps

IngaSokman:
Ready for review

src/router/router.tsx
```diff
@@ -238,6 +244,10 @@ const router = createBrowserRouter([
     path: 'test',
     element: <Components />,
   },
+  {
```
superspartak: Probably shouldn't be here

superspartak (approved):
Seems good!

VaabelMaret (approved):

PR #100, opened 2023-09-29
Improve-428-added shortname ja name validation and fixed styling

VaabelMaret:


IngaSokman (commented):
Screenshot 2023-10-02 at 15 32 17
Getting console error

IngaSokman:
**[blocker]** About "name" and "short name" - in WIKI there is said "short name has to be text only", but currently can add numbers or characters to both of the names

IngaSokman (changes requested):
**[blocker]** About "name" and "short name" - in WIKI there is said "short name has to be text only", but currently can add numbers or characters to both of the names

IngaSokman (approved):

PR #102, opened 2023-10-04
NEW-303-added breadcrumbs component and routes

VaabelMaret:
NOTE: single vendors fetch is not fixed yet, but will be after BE deploy


src/components/molecules/Breadcrumbs/BreadcrumbsTitle.tsx
```diff
@@ -0,0 +1,45 @@
...
+}: BreadcrumbComponentProps<ParamKey>) => {
+  const { t } = useTranslation()
+
+  const { vendorId, userId, orderId }: idTypes = match?.params
+
+  const { vendor } = useVendorFetch({ id: vendorId })
+  const { user } = useFetchUser({ id: userId })
+
+  const { name } = useMemo(() => {
+    switch (true) {
```
superspartak, line 29: It is a little bit unconventional. Is there a reason not to go with  a classic if else in this case?

VaabelMaret: Yes, I have an approach that if there is 3 or more conditions then I prefer use switch, because this is more readable.
Or did you think something else in here?

VaabelMaret:
single vendor fetch is done now

VaabelMaret:
NOTE! Not to merge yet, there is some pages missing

superspartak (approved):

PR #105, opened 2023-10-09
347

superspartak:


src/components/molecules/TaskCandidatesSection/TaskCandidatesSection.tsx
```diff
@@ -67,15 +74,18 @@ const TaskCandidatesSection: FC<TaskCandidatesSectionProps> = ({
...
       }),
     [candidates, getCandidateStatus]
   )
 
-  const handleDelete = useCallback((candidateId: string) => {
-    // TODO: remove candidate with id: candidateId from this task
-  }, [])
+  const handleDelete = useCallback(
+    (vendor_id: string) => {
+      deleteAssignmentVendor({ data: [{ vendor_id }] })
```
VaabelMaret: [blocker]
Deleting error/success message is missing

src/components/organisms/tables/SelectVendorsTable/SelectVendorsTable.tsx
```diff
@@ -88,14 +88,19 @@ const SelectVendorsTable = <TFormValues extends FieldValues>({
   handleSortingChange,
   handlePaginationChange,
```
VaabelMaret, line 89: [blocker]

filtering and sorting not working.
Skills are missing, but filtering works.

VaabelMaret: [blocker]

In 'FreeDaysHearder' if we don't have info text then info icon should be hidden

VaabelMaret (changes requested):

IngaSokman:
[blocker] (maybe?) Noticed that when I have already selected a vendor from the list, and then went back to adding vendor modal, cleared filters, then the person that I previously selected, the checkbox is not ticked for that vendor

IngaSokman:
[blocker] On the modal, when I press "Kuvan korraga" button, nothing happens

IngaSokman:
[non-blocker] Getting eslint error in useUsers file, just format it :D

src/components/molecules/TaskCandidatesSection/TaskCandidatesSection.tsx
superspartak: Do we use showNotification for that? Is there a specific message described anywhere?

VaabelMaret: yes, we use showNotification. we don't have the messages, but make them all similar.

superspartak:
> [blocker] On the modal, when I press "Kuvan korraga" button, nothing happens

BE related, being fixed

superspartak:
> [blocker] (maybe?) Noticed that when I have already selected a vendor from the list, and then went back to adding vendor modal, cleared filters, then the person that I previously selected, the checkbox is not ticked for that vendor

Good point. What I understood from previous discussions is that this modal is used only to add vendors. If the added vendors are pre-selected, it would give the impression that you can remove them by un-selecting the vendor imo. But the modal is only for adding. If the same vendor is selected again and were tried to be added, it shows the user an error about that. So I would say it works as intended.

src/components/molecules/TaskCandidatesSection/TaskCandidatesSection.tsx
superspartak: Done, please check

src/components/organisms/tables/SelectVendorsTable/SelectVendorsTable.tsx
superspartak: FreeDaysHearder proper information added, thx to Birgit.

Filtering on languages now works. Skills are added. When you say sorting not working, what do you mean? Which column does not work?

IngaSokman (approved):

PR #106, opened 2023-10-09
New-471 Fix typo, fix column width

IngaSokman:


superspartak (approved):
Seems OK. But maybe get some extra opinions regarding a better way to handle the column information that is now hidden in overflow

IngaSokman:
@superspartak Discussed it just with MJ, he said that let's keep it currently hidden, if designer or client desides someting else, I will change it

VaabelMaret (approved):

PR #117, opened 2023-10-15
385

IngaSokman:


src/hooks/requests/useOrders.ts
```diff
@@ -197,3 +197,20 @@ export const useSubOrderWorkflow = ({ id }: { id?: string }) => {
     isLoading,
   }
 }
+
+export const useToggleMtEngine = ({ id }: { id?: string }) => {
```
superspartak: Id of what exactly? Feature? Assignment? Order? Can be looked up from swagger indeed but a little bit of a more descriptive name when it comes to "id" would be better 😄

superspartak (approved):

PR #119, opened 2023-10-24
Improve-495-fixed tm privileges and removed empty tags and added fixe…

VaabelMaret:
…d size to tm details row


src/components/organisms/tables/TranslationMemoriesTable/TranslationMemoriesTable.tsx
```diff
@@ -130,7 +126,7 @@ const TranslationMemoriesTable: FC = () => {
           <div className={classes.tagsRow}>
             {map(getValue(), (value) => {
               const tagName = find(tagsOptions, { value })?.label || ''
-              return <Tag label={tagName} value key={value} />
+              return !!tagName && <Tag label={tagName} value key={value} />
```
superspartak, line 129: Maybe {tagName ? <Tag label={tagName} value key={value} /> : null} can be better, since that is the standard doc recommendation

PR #120, opened 2023-10-24
WIP-524- FE updated to reflect the changes made in API regarding the …

superspartak:
…features values being injected to assignments in subproject data

### DIFF
--- a/src/components/molecules/Assignment/Assignment.tsx
+++ b/src/components/molecules/Assignment/Assignment.tsx
@@ -51,8 +51,8 @@ const Assignment: FC<AssignmentProps> = ({
   candidates,
   id,
   assigned_vendor_id,
-  assignee_id,
-  feature,
+  assignee,
+  job_definition,
   source_language_classifier_value_id,
   destination_language_classifier_value_id,
   projectDeadline,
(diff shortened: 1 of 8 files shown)


src/components/templates/SubOrderSection/SubOrderSection.tsx
```diff
@@ -133,20 +125,18 @@ const SubOrderSection: FC<SubOrderProps> = ({
...
-      correspondingAssignments,
-      ({ assigned_vendor_id }) => !assigned_vendor_id
-    )
-  })
+  const hasAnyUnassignedFeatures = !find(
+    assignments,
+    ({ assignee }) => !assignee
+  )
+
+  const tabs = assignments.map(
```
IngaSokman: For consistency reasons would be nice to use already imported map from lodash

IngaSokman (approved):

VaabelMaret:
[blocker]
catSupported now comes from a different place. `project.type_classifier_value.project_type_config.cat_tool_enabled`

src/components/molecules/Assignment/Assignment.tsx
```diff
@@ -79,6 +79,8 @@ const Assignment: FC<AssignmentProps> = ({
     [vendor]
   ) as DiscountPercentages
 
+  const feature = job_definition.job_key
```
VaabelMaret, line 82: If there isn't any job_key,  this part might give as an error

VaabelMaret (changes requested):

src/components/molecules/Assignment/Assignment.tsx
superspartak: Dima confirmed that it can never be null

VaabelMaret (approved):

PR #124, opened 2023-10-25
525 order 

VaabelMaret:
This contains also 509


src/components/organisms/tables/SubOrdersTable/SubOrdersTable.tsx
```diff
@@ -107,7 +107,12 @@ const SubOrdersTable: FC = () => {
   const onSubmit: SubmitHandler<FormValues> = useCallback(
     (payload) => {
       handleFilterChange({
```
superspartak, line 109: handleFilterChange({
         ...payload,
           only_show_personal_projects: payload?.only_show_personal_projects ? 1 : 0
       })
       
Might be easier on the eye

PR #135, opened 2023-10-28
Improve-383-added dealine and comment update logic to assignment

VaabelMaret:


src/hooks/requests/useAssignments.ts
```diff
@@ -213,26 +213,37 @@ export const useAssignmentRemoveVolume = ({
...
-    // },
+      apiClient.put(`${endpoints.ASSIGNMENTS}/${id}`, payload),
+    onSuccess: ({ data }: { data: AssignmentType }) => {
+      queryClient.setQueryData(
+        ['suborders', data.sub_project_id],
+        (oldData?: SubOrderResponse) => {
+          const { data: previousData } = oldData || {}
+          if (!previousData) return oldData
+
+          const newAssignments = previousData.assignments.map((item) => {
```
superspartak: [non-blocker] use lodash map

src/hooks/requests/useAssignments.ts
```diff
@@ -213,26 +213,37 @@ export const useAssignmentRemoveVolume = ({
 }
 
 export const useAssignmentUpdate = ({ id }: { id?: string }) => {
-  // const queryClient = useQueryClient()
+  const queryClient = useQueryClient()
   const { mutateAsync: updateAssignment, isLoading } = useMutation({
-    mutationKey: ['assignments', id],
+    mutationKey: ['subOrders', id],
```
superspartak: is O being uppercase intentional?

src/components/molecules/DateTimePicker/DateTimePicker.tsx
```diff
@@ -32,9 +33,12 @@ const DateTimePicker = forwardRef<HTMLInputElement, DateTimePickerProps>(
       className,
       minDate,
       maxDate,
+      onDateTimeChange,
     } = props
 
     const { t } = useTranslation()
+    const [isModalOpen, setIsModalOpen] = useState<boolean>(false)
+    const [wasModalOpen, setWasModalOpen] = useState(false)
```
superspartak, line 39: Since we don't use wasModalOpen to trigger a re-render but rather just to check a boolean value, maybe it might be better to convert it into a ref instead?

src/components/molecules/Assignment/Assignment.tsx
```diff
@@ -167,9 +196,74 @@ const Assignment: FC<AssignmentProps> = ({
...
+        handleUpdateAssignment({
+          deadline_at: formattedDateTime,
+        })
+      }
+    },
+
+    [formattedDeadline, handleUpdateAssignment]
+  )
+
+  useEffect(() => {
```
superspartak: BLOCKER
There is a problem here. During the first mounting of the component, textAreaElement points to the <body>, and therefore does not update when used directly after opening the page. Only after a state re-render (for example, by changing the deadline) does it get updated to the correct text input element.

src/components/molecules/Assignment/Assignment.tsx
```diff
@@ -167,9 +196,74 @@ const Assignment: FC<AssignmentProps> = ({
     }
   }, [t])
 
-  // TODO: shouldShowStartTimeFields no info about where to take this from yet
-  const shouldShowStartTimeFields =
-    project?.type_classifier_value?.project_type_config?.is_start_date_supported
+  const handleUpdateAssignment = useCallback(
```
superspartak, line 199: [BLOCKER]
Another thing (probably caused by the onSuccess callback of useAssignmentUpdate) is that it changes the order of assignments. After every update, the updated assignment goes to the bottom. So Teostaja 3 seems to have the assignment that should be the first in the order. Rather than adding the updated data to the end of array, update it in place.

src/components/molecules/DateTimePicker/DateTimePicker.tsx
VaabelMaret: This worked the best in this situation

IngaSokman (approved):

src/components/molecules/DateTimePicker/DateTimePicker.tsx
KaarelKa: Yeah usually in cases where you want to see the previous value of some prop or state variable (isModalOpen) in this case, we use a ref. ahooks actually has a hook for this as well called usePrevious. Not much of an issue here, usually just better for avoiding unnecessary rerendering + option to avoid duplication of state changes.

src/hooks/requests/useAssignments.ts
```diff
@@ -215,26 +215,17 @@ export const useAssignmentRemoveVolume = ({
...
-    //       if (!previousData) return oldData
-    //       const newData = { ...previousData, ...data }
-    //       return { data: newData }
-    //     }
-    //   )
-    //   queryClient.refetchQueries({ queryKey: ['institutions'], type: 'active' })
-    // },
+      apiClient.put(`${endpoints.ASSIGNMENTS}/${id}`, payload),
+    onSuccess: () => {
+      queryClient.refetchQueries({
```
KaarelKa, line 224: **[non-blocker]** Hmm do we need the refetch here ? Ideally would prefer to avoid unnecessary requests. As far as I know the only thing that changes here is 1 assignment of 1 suborder and that assignment is returned as a response from BE, so we should be able to just switch out the corresponding assignment in that suborder instead of refetching all suborders.

But this is not a big problem right now and can be skipped to save some time

src/components/molecules/TextInput/TextInput.tsx
```diff
@@ -101,6 +103,9 @@ const TextInput = forwardRef<
             <textarea
               {...(inputProps as unknown as InputHTMLAttributes<HTMLTextAreaElement>)}
               rows={4}
+              onBlur={() =>
```
KaarelKa: **[blocker]**

This is a little dangerous. onBlur is also passed from react-hook-form for every input to make sure values are saved into the form in every possible scenario. With this you are overwriting the onBlur passed from react-hook-form (it is inside the "inputProps" object here).

KaarelKa (changes requested):

src/components/molecules/TextInput/TextInput.tsx
VaabelMaret: Fixed

PR #138, opened 2023-11-06
538 - Fix form not resetting on cancellation and search inputs going over the edge

rebekkabb:
Ticket - https://github.com/keeleinstituut/tv-tolkevarav/issues/538

To-do:
- Fix data not resetting when we edit an order and press the cancel button. The changes you made should not be seen after you cancel
- Fix the search inputs inside the drop-down elements in the order editing/adding going over the edge

How to test - Edit an order then press cancel and check if the edited info has disappeared and check the drop-down searches if they look contained. Also check the order adding view for the data reset and search input containment as well.


IngaSokman (changes requested):
1. "Tellimuse sildid" dropdown searchbox is overflowing
2. The editing of the existing order does not work quite properly. Went over flow (added text for field "Erijuhised tellimuse kohta", when starting the text was "Test1" for the field): 1. Changed text to "Test2", 2. Changed text again to "Test 3", 3. Started to edit the text again, but pressed "Cancel" instead of "Save" - the field shows me wrong text - "Test1"

KaarelKa (approved):

PR #139, opened 2023-11-06
Improve-535-added order event-start-at types

VaabelMaret:


VaabelMaret:
Note: can't test currently, BE needs add the support

src/types/orders.ts
```diff
@@ -47,6 +47,13 @@ export enum WorkflowTemplateID {
   SampleProject = 'Sample-project',
 }
 
+export enum TypesWithStartTime {
+  ORAL_TRANSLATION = 'ORAL_TRANSLATION',
```
KaarelKa: **[blocker]**

Would be good to use PascalCase for these enum keys for consistency

KaarelKa (changes requested):

src/types/orders.ts
VaabelMaret: fixed

IngaSokman (approved):

KaarelKa (approved):

PR #144, opened 2023-11-07
501 - Adjust landing page buttons and redirection

rebekkabb:
Ticket - https://github.com/keeleinstituut/tv-tolkevarav/issues/501

To-do: 
- Make "Lisa uus tellimus" button text not be on multiple lines
- Make clicking on "Vaata kasutusjuhendit" button direct the user to the user manual page
- Remove "Vaata kasutusjuhendit" button from the unauthorized landing page, there should only be "Sisene" button

How to test - Check both of the landing pages (logged in and not logged in), click on the buttons to verify their intended behaviour

### DIFF
--- a/src/components/organisms/LandingContent/LandingContent.tsx
+++ b/src/components/organisms/LandingContent/LandingContent.tsx
@@ -27,6 +27,7 @@ const LandingContent: FC<LandingContentProps> = ({ className }) => {
             children={t('button.check_tutorial')}
             href={isUserLoggedIn ? '/manual' : undefined}
             className={classes.bigButton}
+            hidden={!isUserLoggedIn}
           />
           <Button
             appearance={AppearanceTypes.Primary}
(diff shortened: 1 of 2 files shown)


src/components/organisms/LandingContent/LandingContent.tsx
```diff
@@ -25,7 +25,9 @@ const LandingContent: FC<LandingContentProps> = ({ className }) => {
           <Button
             appearance={AppearanceTypes.Secondary}
             children={t('button.check_tutorial')}
+            href={isUserLoggedIn ? '/manual' : undefined}
```
KaarelKa, line 28: **[non-blocker]**

Not too important, but the condition here is unnecessary. Can be just href="/manual" since we won't show the button if it's not manual anyway (if user is not logged in)

KaarelKa (approved):

PR #150, opened 2023-11-10
351

KaarelKa:
Removed restrictions of adding volume.
Note* Currently you will get a BE error, but the restriction will also be removed from BE soon


VaabelMaret (approved):

IngaSokman (approved):

rebekkabb (approved):
lgtm

PR #151, opened 2023-11-13
Fix-553 - Order table filter and dropdown changes

rebekkabb:
Ticket - https://github.com/keeleinstituut/tv-tolkevarav/issues/553

To-do: 
- Add missing project type translations to the table
- Add the right drop-down and filtering to "Keelesuunad" and "Tellimuse sildid"
- Fix "Keelesuunad" filter resetting when an additional filter is picked after it. Issue appears in "Tellimused" and "Teostajate andmebaas" tables.

How to test - 
Go to the "Tellimused" view and look if the types are translated and try to filter by "Tellimuse sildid" and "Keelesuunad", try using both filters at the same time as well. 
Also check out "Teostajate andmebaas" table filters, because this table also had the third todo issue.

### DIFF
--- a/src/components/organisms/tables/OrdersTable/OrdersTable.tsx
+++ b/src/components/organisms/tables/OrdersTable/OrdersTable.tsx
@@ -92,7 +92,7 @@ const OrdersTable: FC = () => {
             ext_id,
             reference_number,
             deadline_at,
-            type: type_classifier_value?.value || '',
+            type: type_classifier_value?.name || '',
             status,
             tags: map(tags, (value) => value?.name || ''),
             price,
--- a/src/components/organisms/tables/OrdersTable/OrdersTable.tsx
+++ b/src/components/organisms/tables/OrdersTable/OrdersTable.tsx
@@ -125,23 +125,26 @@ const OrdersTable: FC = () => {
 
   const handleModifiedFilterChange = useCallback(
     (filters?: FilterFunctionType) => {
-      const { language_directions, ...rest } = filters || {}
-      const typedLanguageDirection = language_directions as string[]
+      let currentFilters = filters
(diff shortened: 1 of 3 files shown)


src/components/organisms/tables/OrdersTable/OrdersTable.tsx
```diff
@@ -119,6 +123,33 @@ const OrdersTable: FC = () => {
     },
   })
 
+  const handleModifiedFilterChange = useCallback(
+    (filters?: FilterFunctionType) => {
+      let currentFilters = filters
```
KaarelKa, line 128: **[non-blocker]**

Not too important, but would suggest avoiding any data mutations, when dealing with React projects. They can have some unintended side effects. This is mostly related to anything stored to state and/or used by react for rerendering. It is fine to use in some cases, but usually we try to avoid doing it in React projects, to make sure we don't end up with some unintended behaviour + to reduce the chance that people will use that code as example in some other context, where it will cause issues.

**Note**: It doesn't cause any issues here

KaarelKa (approved):

src/components/organisms/tables/OrdersTable/OrdersTable.tsx
rebekkabb: Noted! I thought the code would be easier to read if I didn't call the `if (handleFilterChange) { handleFilterChange(...) }` twice with the two possible options separately.

KaarelKa: There is another option to do sth like 


const currentFilters = {
    ...rest,
    ...(isEmpty(modifiedLanguageDirections)
      ? { language_directions: modifiedLanguageDirections }
      : {}),
}

  
  But yeah, not really important, can stay the way it is

PR #160, opened 2023-11-15
Improve - 69 - Add order cancellation confirmation modal + integrate …

KaarelKa:
…with BE

Task: https://github.com/keeleinstituut/tv-tolkevarav/issues/244


KaarelKa:
**Note*** Be currently doesn't accept the "reason" and "comments" fields. This part might need to be changed, once BE part is implemented

PR #163, opened 2023-11-16
503 - Institutions Price Rate Percentages improvement

rebekkabb:
Ticket - https://github.com/keeleinstituut/tv-tolkevarav/issues/503

To-do
- In "Tehnilised sätted" make the "Tühista" button disabled until user edits any field. After clicking the "Tühista" button, the fields should reset to the state that they were before the edit. Furthermore when the user edits something, the save button turns blue. If user changes back to how it was before editing, the save button will turn back grey.
- Validation errors not appearing on touch in "Teostajate Andmebaas" vendor editing and in "Kasutajate haldus" user editing
- User has the privilege: "Asutuse hinnakirja vaatamine" (view_institution_price_rate), no action buttons should be shown

How to test 
- Change form values and try cancelling, submitting, check for validation errors in "Tehnilised sätted"
- Check the forms for validation errors on touch in "Teostajate Andmebaas" and "Kasutajate haldus"
- Check "Tehnilised sätted" form with "Asutuse hinnakirja vaatamine" (view_institution_price_rate) privilege. Use Stephanie Hanna 43608102253 user for example.


IngaSokman (approved):
Add validation fix on "Kasutajate haldus" for phone number, otherwise approved

src/pages/TechnicalSettings/TechnicalSettings.tsx
```diff
@@ -22,6 +22,36 @@ const TechnicalSettings: FC = () => {
   const { institutionDiscounts } = useFetchInstitutionDiscounts()
   const { updateInstitutionDiscounts } = useUpdateInstitutionDiscounts()
 
+  const defaultValues = useMemo(
```
KaarelKa, line 25: **[non-blocker]**

Quite a lot of repetition here, would suggest using something to map these from institutionDiscounts, to avoid repetition. For example with lodash mapValues it would be:

const defaultValues = mapValues(institutionDiscounts, (value) => String(value) || '0')

KaarelKa (approved):

src/pages/TechnicalSettings/TechnicalSettings.tsx
rebekkabb: Ooh fancy, will use that 👍

KaarelKa (approved):

PR #164, opened 2023-11-16
173 audit logs

VaabelMaret:


IngaSokman (approved):

rebekkabb (approved):
Overall looks nice to me, found a couple of discussion points/non-blockers:

- If you look at the details for log type that has many details like "Objekti muutmine" then you can get this error in the console: `Warning: Each child in a list should have a unique "key" prop.`

- Contents of the Filtrid box also overflows like the table on smaller screens

- Filtering is shifted by 2 hours - on discussion seems to be issue with data sent from backend

- The form buttons "Puhasta väljad" and "Filtreeri" are quite small, bigger buttons would be nicer

- Should the import CSV button be clickable when no filters are chosen? Currently returns csv with all? the logs

src/components/molecules/InputWrapper/InputWrapper.tsx
```diff
@@ -29,14 +30,20 @@ const InputWrapper = forwardRef<HTMLInputElement, InputWrapperProps>(
       wrapperClass,
       errorClass,
       errorZIndex,
+      paginationLabelClassName,
```
KaarelKa: **[bloker]**

Do we need this ? We already have the option to pass "className" prop. If you want to overwrite any styles from the "classes.container" then you can do so inside the class passed as "className". It might be slightly more css for this 1 case, but if we start adding extra props to a commonly used componend for every usecase, we will end up with something that will be quite unmaintainable in the future. I would pick maintainability + readability over less code in this case.

src/components/molecules/TimePickerInput/TimePickerInput.tsx
```diff
@@ -82,12 +85,22 @@ const TimeInput = forwardRef<HTMLInputElement, TimeInputProps>(
             placeholder: '0',
           })}
         />
-        <Clock
-          className={classNames(
-            classes.timeIcon,
-            disabled && classes.disabledIcon
-          )}
-        />
+        {icon ? (
```
KaarelKa: **[non-blocker]**

Could make this shorter + avoid inline conditional rendering by moving the conditional part to the "icon" prop passed here:

   <Icon
            icon={icon || Clock}
            className={classNames(
              classes.timeIcon,
              disabled && classes.disabledIcon
            )}
          />

src/components/molecules/ToggleTabs/ToggleTabs.tsx
```diff
@@ -23,21 +25,36 @@ const ToggleTabs: FC<ToggleTabsProps> = ({
   setActiveTab,
   tabs = [],
   hidden,
+  dateTabsClassName,
+  value,
```
KaarelKa, line 25: **[blocker]**

A bit confused about these 2 props (value and onChange), they seem to just duplicate the exact behaviour of activeTab and setActiveTab without doing anything extra. If you just want to rename the original ones, so you could us the tabs as a Form Input component, then let's remove the previous names for these props rather than duplicate them

src/components/organisms/tables/LogsSubRowTable/LogsSubRowTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+  TableSizeTypes,
+} from 'components/organisms/DataTable/DataTable'
+import { createColumnHelper, ColumnDef } from '@tanstack/react-table'
+import classes from './classes.module.scss'
+import { useTranslation } from 'react-i18next'
+import classNames from 'classnames'
+import { includes, join, map, omit, pickBy, toLower } from 'lodash'
+import { EventParameters, ObjectParameters } from 'types/auditLogs'
+
+type AuditLog = {
```
KaarelKa: **[non-blocker]**

There seem to be at least 2 duplicates for AuditLog type, kind of even 3. Would be good to have just 1, where the actual structure of the AuditLog is defined and in other places you can use Omit or Pick to take only the fields you need from there

src/hooks/requests/useAuditLogs.ts
```diff
@@ -0,0 +1,61 @@
...
+      })
+    },
+  })
+  return {
+    isLoading,
+    exportCSV,
+  }
+}
+
+export const useEventTypesFetch = () => {
```
KaarelKa: **[blocker]**

Should be under hooks and renamed, since this is not fetching anything and not related to API requests at all. Personally I wouldn't even create a helper for it, but rather map it where it's needed 

      options: map(EventTypes, (type) => ({
        value: type,
        label: i18n.t(`logs.event_type.${type}`),
      })),
      
But if you want to create a hook out of it, then it should just be under /hooks folder

KaarelKa: Extra note: avoid importing i18n directly and using it like this, except for cases, when there is no other option. the "t" function should always be used from the hook, to make sure the correct translations are taken, when the pages language is changed. importing i18n directly and using the "t" function from there won't update right away and will cause issues. Once we add actual language switching we will need to go over the places where it is imported directly and make sure that they will refresh correctly, so we want to keep those places to a minimum

src/pages/Logs/Logs.tsx
```diff
@@ -1,9 +1,300 @@
...
+
+      <Container className={classes.container}>
+        <div className={classes.buttonsContainer}>
+          <h4 className={classes.filters}>{t('logs.filters')}</h4>
+          <div className={classes.buttonsContainer}>
+            <Button
+              children={t('logs.clean_fields')}
+              ariaLabel={t('logs.clean_fields')}
+              size={SizeTypes.S}
+              onClick={() => reset()}
```
KaarelKa: or just onClick={reset}

KaarelKa (changes requested):

src/components/molecules/InputWrapper/InputWrapper.tsx
VaabelMaret: no need

src/pages/Logs/Logs.tsx
VaabelMaret: I usually use shorter version, but in that case this will give me the typescript error

VaabelMaret:
ready to review

src/pages/Logs/Logs.tsx
KaarelKa: Okay true yeah, reset doesn't match the typing we have for onClick. Ideal solution would be to change the onClick typing a bit to allow for it, but let's keep it as is for now, it's not really important

KaarelKa (approved):

KaarelKa:
Hmm doesn't seem to match design, when I change the screen width to 1440 (width in design). Wrapping to second line with forms is fine, if it happens on smaller screens, but ideally at least 1440px+ width screens should still match the design

PR #168, opened 2023-11-20
FIX-511 - Add pagination index reset logic during filtering

rebekkabb:
Ticket - https://github.com/keeleinstituut/tv-tolkevarav/issues/511

To-do - If user is on page 3 or further and the search result happens to have less pages then the current page, the user cannot see the search results without clicking on the navigation. This happens on all tables.

How to test - For example go to "Alam-tellimused" and click on pagination bigger than 2 and type `10--5` into the search, the pagination should now reset to page one so you can see the one result. Can also test all the other tables for the same.


KaarelKa:
Looks good, but we should maybe reset the pagination whenever search (or any other filter) changes. I think this is how it usually works, since pagination will be specific to the applied filters

PR #178, opened 2023-11-24
511 - Improve pagination reset during filter change

rebekkabb:
It was previously mentioned in 511 that the pagination should reset whenever the filter changes. 

Did some cleanup and fixed a couple of filter and pagination bugs: 
- Suborder statuses filter not working
- Pagination calculations going into minus when last page is 4


src/components/organisms/DataTable/DataTable.tsx
```diff
@@ -105,24 +108,14 @@ const DataTable = <TData,>(
...
     subRowComponent,
+    pagination,
+    setPagination,
   }: DataTableProps<TData>,
   ref: Ref<HTMLDivElement>
 ) => {
   const [horizontalWrapperId] = useState(useId())
   const { per_page, last_page } = paginationData || {}
   const [expanded, setExpanded] = useState<ExpandedState>({})
-  const [pagination, setPagination] = useState<PaginationState>({
```
KaarelKa, line 114: **[blocker]** (Possibly)

I would prefer a less "intrusive" solution, where we change far less things. The general logic is that if filters change, then reset pagination. As the pagination data for every BE request comes from/uses our useFilters hook, it would be a very good place where to handle it. 

Basically inside the "handleFilterChange" and "handleSortingChange" functions we could just call handlePaginationChange({ page: 0 }) (since the per_page value shouldn't change). And we would get away with essentially 2 added lines in 1 file + Possibly handling this manually, if we have any exceptions somewhere, that don't use the useFilters helper hook

KaarelKa (changes requested):

src/components/organisms/DataTable/DataTable.tsx
rebekkabb: Couldn't use handlePaginationChange there since the handle methods seem to overwrite each other, but I found a workaround and the overall result is definitely much less intrusive. Take look when you can and let me know what you think!

KaarelKa (approved):

PR #179, opened 2023-11-28
578

KaarelKa:


KaarelKa:
Also contains changes for [454](https://github.com/keeleinstituut/tv-tolkevarav/issues/454) task

src/hooks/requests/useVolumes.ts
```diff
@@ -0,0 +1,210 @@
...
+  const { mutateAsync: removeAssignmentVolume, isLoading } = useMutation({
+    mutationKey: ['subprojects', id],
+    mutationFn: (payload: { volumeId?: string }) =>
+      apiClient.delete(`${endpoints.VOLUMES}/${payload.volumeId}`),
+    onSuccess: (_, { volumeId }) => {
+      queryClient.setQueryData(
+        ['subprojects', id],
+        (oldData?: SubProjectResponse) => {
+          const { data: previousData } = oldData || {}
+          if (!previousData) return oldData
```
rebekkabb: Not a blocker but maybe also extract the logic here into a separate method like getNewSubProjectWithAssignment

rebekkabb (approved):
Left one comment, but the modal seems to work and the other stuff lgtm

src/hooks/requests/useVolumes.ts
KaarelKa: Yeah not a bad idea, I usually do it once I have at least 2 places where I need to use something, but could be done before here as well

PR #180, opened 2023-11-28
390

IngaSokman:
Adding comments and category is not yet added on BE side.


KaarelKa:
Let's use the multi-select dropdowns for table header filters. Main issue is that with single select the UX is really bad. Can't see what I have selected + can't clear the selection, which makes the filters very hard to use. It seems that BE also supports multiple options + it would be more consistent compared to other table filters

src/pages/MyTasks/MyTasks.tsx
```diff
@@ -1,67 +1,150 @@
...
+  handleFilterChange: (value?: FilterFunctionType) => void
+  handleSortingChange: (value?: SortingFunctionType) => void
+  handlePaginationChange: (value?: PaginationFunctionType) => void
+  isLoading: boolean
+  isHistoryTab?: boolean
+}
 
-interface ObjectType {
-  [key: string]: string
+const FetchTasks = (assignedToMe: number) => {
```
KaarelKa: **[non-blocker]**

This seems unnecessary, just a wrapper around useFetchTasks for renaming the parameter. I would suggest using useFetchTasks directly to reduce complexity

src/pages/MyTasks/MyTasks.tsx
```diff
@@ -1,67 +1,150 @@
...
+    handleFilterChange: handleHistoryFilterChange,
+    handleSortingChange: handleHistorySortingChange,
+    handlePaginationChange: handleHisoryPaginationChange,
+  } = useFetchHistoryTasks()
+
+  const [activeTab, setActiveTab] = useState<string | undefined>(
+    t('my_tasks.my_assignments')
+  )
+
+  let Component: FC<TasksTableProps> = () => null
```
KaarelKa: **[blocker]**

This approach is usually used, when we the Component can be at least 2 different things, otherwise it doesn't make much sense. In this case I see that the "Component" is always "TasksTable" so would just render <TasksTable /> and have some variable for the props we pass in, since those are the only things that change based on activeTab.

src/pages/MyTasks/MyTasks.tsx
```diff
@@ -1,67 +1,150 @@
...
+        handleSortingChange: handleHistorySortingChange,
+        handlePaginationChange: handleHisoryPaginationChange,
+        isHistoryTab: true,
+      }
+      break
+    default:
       break
   }
 
+  const taskTabs = chain([
```
KaarelKa: **[non-blocker]**

chain is a good helper function, though i would probably not use it, if I only want to call 2 functions, normally it's used when there are far more functions to call in a row. 

**Extra note:** This doesn't matter in this project, since we are not dealing with SEO, but there is 1 downside with chain that applies in any project which needs good SEO. The issue is that with "chain" you import every function from lodash, which will bloat the size of any page that uses it

src/hooks/requests/useTasks.ts
```diff
@@ -37,6 +37,65 @@ export const useFetchTasks = (initialFilters?: TasksPayloadType) => {
   }
 }
 
+export const useFetchTask = ({ id }: { id?: string }) => {
```
KaarelKa, line 41: **[blocker]**

enabled should be false, when there is no id, otherwise we get spammed with errors on a lot of pages, where BreadcrumbsTitle is calling this hook, but not passing any id. Same goes for useFetchHistoryTasks

src/components/molecules/ExpandableContentContainer/ExpandableContentContainer.tsx
```diff
@@ -18,6 +18,7 @@ interface ExpandableContentContainerProps {
   onExpandedChange?: (isExpanded: boolean) => void
   id?: string
   initialIsExpanded?: boolean
+  isExpanedDisabled?: boolean
```
KaarelKa: small typo here isExpandedDisabled*

src/components/molecules/FinalFilesList/FinalFilesList.tsx
```diff
@@ -160,59 +174,105 @@ const FinalFilesList = <TFormValues extends FieldValues>({
   }, [sharedFiles, typedValue])
 
   const columns = [
-    columnHelper.accessor('shared_with_client', {
-      header: () => <span>{t('label.shared_with_client')}</span>,
+    ...(mode !== 'view'
```
KaarelKa: **[blocker]**

With enums we should always use them when comparing as well. So "mode !== ProjectDetailModes.View" . The main point behind enums is to avoid all typos, so we should use it fully. (There are some other places in this file with the same issue)

src/components/molecules/FinalFilesList/FinalFilesList.tsx
```diff
@@ -224,6 +284,34 @@ const FinalFilesList = <TFormValues extends FieldValues>({
       : []),
   ] as ColumnDef<FileRow>[]
 
+  // TODO: possibly not needed
```
KaarelKa: **[blocker]** (Possibly)

I don't think we need this

src/components/molecules/SourceFilesList/SourceFilesList.tsx
```diff
@@ -125,7 +136,8 @@ const SourceFilesList = <TFormValues extends FieldValues>({
   )
 
   const columns = [
-    ...(canGenerateProject && isEditable
+    ...((canGenerateProject && isEditable) ||
+    (canGenerateProject && mode !== 'view')
```
KaarelKa: **[blocker]**

Mentioned above, but should use enums correctly

src/components/organisms/tables/TasksTable/TasksTable.tsx
```diff
@@ -1,302 +1,238 @@
...
-            //       sub_projects,
-            //       ({
-            //         source_language_classifier_value,
-            //         destination_language_classifier_value,
-            //       }) =>
-            //         `${source_language_classifier_value?.value} > ${destination_language_classifier_value?.value}`
-            //     )
-            //   ),
-          }
+  const tasksData = useMemo(() => {
```
KaarelKa, line 67: **[blocker]**

It seems that some tasks are getting filtered out here, which should be visible. For example all "CLIENT_REVIEW" tasks

src/pages/TaskPage/TaskPage.tsx
```diff
@@ -1,386 +1,116 @@
...
+    ? historyTask || {}
+    : task || {}
+
+  const { subProject, ext_id, sub_project_id, volumes, comments } =
+    assignment || {}
+
+  const {
+    project,
+    cat_files,
+    source_files,
```
KaarelKa, line 56: **[blocker]** (Possibly)

I'm not sure if we will get the correct source_files from here. I think these are just the source_files of the related subProject, but we should also show the files that were added from the previous task that was completed (Translated files are needed for the Editor for their work). 

**Note**: Okay got an update from Dima about this. Basically it will be source_files + final_files. The final_files are currently not returned here yet, but let's add them anyway, then it will start working, once BE deploys the changes.

src/components/organisms/TaskDetails/TaskDetails.tsx
```diff
@@ -0,0 +1,111 @@
...
+  return (
+    <ExpandableContentContainer
+      className={classNames(
+        classes.expandableContainer,
+        status && classes[toLower(status)]
+      )}
+      onExpandedChange={handleOpenContainer}
+      id={ext_id}
+      isExpanded={!assignee_institution_user_id ? false : isExpanded}
+      rightComponent={<ProjectStatusTag status={status} />}
```
KaarelKa: **[blocker]**

We should be using the status of the subProject here with the jobName (Same as SubProjectSection)

src/components/organisms/TaskContent/TaskContent.tsx
```diff
@@ -0,0 +1,352 @@
...
+import dayjs from 'dayjs'
+import { LanguageClassifierValue } from 'types/classifierValues'
+import BaseButton from 'components/atoms/BaseButton/BaseButton'
+import { ReactComponent as Eye } from 'assets/icons/eye.svg'
+import { apiTypeToKey } from 'components/molecules/AddVolumeInput/AddVolumeInput'
+import { VolumeValue } from 'types/volumes'
+import classNames from 'classnames'
+import Button, { AppearanceTypes } from 'components/molecules/Button/Button'
+import { useCompleteTask } from 'hooks/requests/useTasks'
+import { ProjectDetailModes } from '../ProjectDetails/ProjectDetails'
```
KaarelKa: **[non-blocker]**

Would be good to use absolute imports here as well. Not very important, just future-proofing/consistency

src/components/organisms/TaskContent/TaskContent.tsx
```diff
@@ -0,0 +1,352 @@
...
+  sub_project_id: string
+  volumes?: VolumeValue[]
+  assignee_institution_user_id?: string
+  taskId?: string
+  isHistoryView?: string
+  task_type?: string
+}
+
+const TaskContent: FC<TaskContentProps> = ({
+  deadline_at,
```
KaarelKa, line 57: **[blocker]**

I think we should use the assignment deadline, instead of project deadline here

src/components/organisms/TaskContent/TaskContent.tsx
```diff
@@ -0,0 +1,352 @@
...
+
+  const handleOpenCompleteModal = useCallback(() => {
+    showModal(ModalTypes.ConfirmationModal, {
+      title: t('modal.confirm_complete_task'),
+      modalContent: t('modal.confirm_complete_task_details'),
+      cancelButtonContent: t('button.quit'),
+      proceedButtonContent: t('button.complete'),
+      className: classes.completeModal,
+      handleProceed: handleSubmit(handleCompleteTask),
+      proceedButtonLoading: isCompletingTask,
```
KaarelKa: **[blocker]**

Loading state will never be true. You will open the modal, when isCompletingTask is false and when it changes to true, you are not passing that information to the modal.

src/components/organisms/TaskContent/TaskContent.tsx
```diff
@@ -0,0 +1,352 @@
...
+
+  const handleSendToPreviousAssignmentModal = useCallback(() => {
+    showModal(ModalTypes.ConfirmationModal, {
+      title: t('modal.confirm_send_to_previous_assignment'),
+      modalContent: t('modal.confirm_send_to_previous_assignment_details'),
+      cancelButtonContent: t('button.quit_alt'),
+      proceedButtonContent: t('button.send_back'),
+      className: classes.completeModal,
+      handleProceed: handleSubmit(sendToPreviousAssignment),
+      proceedButtonLoading: isCompletingTask,
```
KaarelKa: **[blocker]**

same issue here

src/components/organisms/TaskContent/TaskContent.tsx
```diff
@@ -0,0 +1,352 @@
...
+        </span>
+        <span
+          className={classNames(
+            classes.taskContainer,
+            !volumes[0] && classes.hideContainer
+          )}
+        >
+          <p className={classes.taskDetails}>{t('label.volume')}</p>
+          <p className={classes.taskContent}>
+            {
```
KaarelKa: **[non-blocker]** I think using the {} wrapper + <> wrapper here are both unnecessary

src/components/organisms/TaskDetails/TaskDetails.tsx
```diff
@@ -0,0 +1,111 @@
...
+      leftComponent={
+        <LeftComponent
+          {...{ ext_id, deadline_at, price, languageDirection }}
+          mode={ProjectDetailModes.View}
+        />
+      }
+      isExpanedDisabled={!assignee_institution_user_id}
+    >
+      <TaskContent
+        deadline_at={deadline_at}
```
KaarelKa: **[blocker]**

I think deadline_at and event_start_at passed here should come from assignment, not project

src/components/organisms/modals/ConfirmationModalBase/ConfirmationModalBase.tsx
```diff
@@ -73,11 +75,13 @@ const ConfirmationModalBase: FC<ConfirmationModalBaseProps> = ({
           onClick: () => {
             if (handleProceed) {
               handleProceed()
+              closeModal()
```
KaarelKa: **[blocker]**

This will break error validation and loading state in every other modal that is built on top of ConfirmationModalBase. This will mean that the modal will close before we can show the loading state or before we get the errors and add them to form inputs inside the modal

src/components/organisms/tables/TasksTable/TasksTable.tsx
```diff
@@ -1,302 +1,238 @@
...
+        filters || {}
+      const typedLanguageDirection = language_direction as string
+      const typedClassifierValueId = type_classifier_value_id as string
 
-  useEffect(() => {
-    // Submit form every time it changes
-    const subscription = watch(() => handleSubmit(onSubmit)())
-    return () => subscription.unsubscribe()
-  }, [handleSubmit, watch])
+      const langPair = {
```
KaarelKa, line 117: **[non-blocker]**

BE should acutally expect an array of object with src and dst languages (same as other endpoints). Currently it'll work with this structure as well, but would be better to stick to arrays. The reason it works like this is more of an exception and not exactly the correct usage

KaarelKa (changes requested):

src/components/organisms/tables/TasksTable/TasksTable.tsx
IngaSokman: Currently waiting on BE answers for adding the missing data

src/components/molecules/FinalFilesList/classes.module.scss
```diff
@@ -114,3 +121,45 @@
     margin-left: auto;
   }
 }
+
+.altFilesContainer {
```
KaarelKa: **[blocker]**

Not used anymore, probably left over after dealing with merge conflicts

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -35,6 +35,7 @@ interface DropdownContentComponentProps extends SelectionControlsInputProps {
   setIsOpen?: Dispatch<SetStateAction<boolean>>
   className?: string
   wrapperRef?: RefObject<HTMLDivElement>
+  isMultiple?: boolean
```
KaarelKa: **[blocker]** (Possibly)

What is the difference between isMultiple and multiple props ? Hard to tell what the functionality would be and which one I should use in the future and what happens if I add both of them or only 1 of them. Right now we have 1 place where both of them are true and a bunch of places where only multiple is true. If we need both of them, then would be good to add some description about what they do and possibly rename 1 of them to better describe it's effect

src/components/organisms/DropdownContent/DropdownContent.tsx
```diff
@@ -293,6 +302,7 @@ const DropdownContent: FC<DropdownContentProps> = ({
         {...rest}
         wrapperRef={wrapperRef}
         ref={clickAwayInputRef}
+        isMultiple={isMultiple}
```
KaarelKa: **[non-blocker]**

It was already passed inside ...rest, I don't think we need to pass it manually

src/components/organisms/TaskContent/TaskContent.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+  const handleOpenCompleteModal = useCallback(() => {
+    showModal(ModalTypes.ConfirmationModal, {
+      title: t('modal.confirm_complete_task'),
+      modalContent: t('modal.confirm_complete_task_details'),
+      cancelButtonContent: t('button.quit'),
+      proceedButtonContent: t('button.complete'),
+      className: classes.completeModal,
+      handleProceed: handleSubmit(handleCompleteTask),
+      proceedButtonLoading: isSubmitting,
```
KaarelKa: **[blocker]**

Same issue exists, you open the modal with isSubmitting: false, and you never update the value inside the modal, so the proceed button will never be in loading state.

src/components/organisms/TaskDetails/TaskDetails.tsx
KaarelKa: **[blocker]**

This issue is unresolved, It still seems to use project status, not subProject status

src/components/organisms/TaskContent/TaskContent.tsx
KaarelKa: **[blocker]**

Issue is not resolved

src/pages/MyTasks/MyTasks.tsx
```diff
@@ -1,67 +1,150 @@
...
+    handleFilterChange: handleHistoryFilterChange,
+    handleSortingChange: handleHistorySortingChange,
+    handlePaginationChange: handleHisoryPaginationChange,
+  } = useFetchHistoryTasks()
+
+  const [activeTab, setActiveTab] = useState<string | undefined>(
+    t('my_tasks.my_assignments')
+  )
+
+  let componentProps: TasksTableProps = {
```
KaarelKa: **[blocker]**

This is basically exactly why we avoid mutating variables/data in React projects. The component will get rendered with these props, but these props can then change inside the "switch" case you have below. This can mean that the component where these props are passed, might not rerender with the new values, but rather with the old ones (reference to "componentProps" will be the same). it will work currently in most cases, because we have quite a lot of unnecessary rerendering here, but this is still very error prone. It will be very hard to validate whether it will rerender correctly every time. 

I would still suggest avoiding mutating arrays/objects and pretty much any other variable in react projects unless you are 100% sure it will work correctly or there is no other way. Current case could easily be solved by sticking useMemo around the switch statement.

const componentProps = useMemo(....

src/types/tasks.ts
```diff
@@ -1,10 +1,27 @@
...
 import {
   PaginationFunctionType,
   ResponseMetaTypes,
   SortingFunctionType,
-} from 'types/collective'
+} from './collective'
+import { ListProject, SourceFile } from './projects'
 import { AssignmentType } from 'types/assignments'
 
+export type SubProject = {
```
KaarelKa: **[blocker]**

Lets avoid duplicated types. If this is meant for getting a subset of fields from subproject type, then there are many typescript helper functions that can help you achieve that (Omit, Pick, Partial for example)

src/components/organisms/TaskContent/TaskContent.tsx
KaarelKa: This is called closure in javascript, you can read a bit more about it in react context from https://adevnadia.medium.com/fantastic-closures-and-how-to-find-them-in-react-d81f000919d2

src/components/organisms/TaskDetails/TaskDetails.tsx
```diff
@@ -0,0 +1,118 @@
...
+      onExpandedChange={handleOpenContainer}
+      id={ext_id}
+      isExpanded={!assignee_institution_user_id ? false : isExpanded}
+      rightComponent={
+        <ProjectStatusTag status={status} jobName={job_short_name} />
+      }
+      wrapContent
+      leftComponent={
+        <LeftComponent
+          {...{ ext_id, deadline_at, price, languageDirection }}
```
KaarelKa, line 101: **[blocker]**

Deadline here should also show the deadline of the assignment, not the entire project. Same goes for price

KaarelKa: **Note*** BE currently doesn't return price of assignment, but it will, so we can add it here already (similarly to final_files).

src/components/organisms/features/GeneralInformationFeature/GeneralInformationFeature.tsx
```diff
@@ -108,7 +108,7 @@ const GeneralInformationFeature: FC<GeneralInformationFeatureProps> = ({
   const defaultValues = useMemo(
     () => ({
       deadline_at: getLocalDateOjectFromUtcDateString(
-        deadline_at || projectDeadlineAt
+        deadline_at || project?.deadline_at || ''
```
KaarelKa: **[blocker]**

These changes are fine, but then it should be applied everywhere. Before we destructured deadline_at from project and renamded it projectDeadlineAt, so you are using the same value here, however in this file the projectDeadlineAt is also still used. Let's just use 1 of them.

src/components/organisms/modals/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -18,6 +18,7 @@ const ConfirmationModal: FC<ConfirmationModalBaseProps> = ({
   return (
     <ConfirmationModalBase
       {...rest}
+      className={className}
```
KaarelKa: **[non-blocker]**

className is not directly used inside this component, so would be better to pass it inside ...rest with all of the other props that are just passed through

src/components/organisms/tables/TasksTable/TasksTable.tsx
```diff
@@ -1,302 +1,244 @@
...
+      }
+
+      if (handleFilterChange) {
+        handleFilterChange(newFilters)
+      }
+    },
+    [handleFilterChange]
+  )
+
+  const columnHelper = createColumnHelper<TaskTableRow>()
```
KaarelKa: **[blocker]**

We need to reduce some unnecessary rerendering here. Won't cause any issues right now, but a couple of things that Maret did for accessibility could be broken by rerendering issues here, even if they aren't then let's still avoid unnecessary rerendering where possible.

To reduce rerendering here, we should use "useMemo" for columns + move the "const columnHelper = createColumnHelper<TaskTableRow>()" outside of the component (Same file just outside of the component, so the columnHelper variable doesn't get remade every time TasksTable rerenders, which in turn would force columns to rerender, so essentially every cell in the table would rerender every time)

src/components/templates/SubProjectSection/SubProjectSection.tsx
```diff
@@ -41,25 +42,36 @@ interface LeftComponentProps {
   ext_id: string
   price?: string
   deadline_at?: string
+  mode?: ProjectDetailModes
 }
 
-const LeftComponent: FC<LeftComponentProps> = ({
+export const LeftComponent: FC<LeftComponentProps> = ({
```
KaarelKa: **[non-blocker]**

Not too important, but usually we only keep these helper components inside the parent component file, if they are not used anywhere else. Once you start using it somewhere else, it should be moved to it's own separate folder as a component (and possibly renamed in this case, since LeftComponent is very hard to understand)

src/types/projects.ts
```diff
@@ -187,30 +187,29 @@ export interface JobDefinition {
 }
 
 export interface ListProject {
-  id: string
-  ext_id: string
-  reference_number: string
-  institution_id: string
-  type_classifier_value_id: string
+  id?: string
```
KaarelKa: **[blocker]** (Possibly)

Some or a lot of these props should always exist for a project (for example id should always exist). It is better to have types that are strict, but what was the use case which required the ? here, so the use case, where some of these fields are missing for a ListProject type object ?

src/types/projects.ts
```diff
@@ -187,30 +187,29 @@ export interface JobDefinition {
...
-  status: ProjectStatus
-  tags: Tag[]
-  price: string
+  status?: ProjectStatus
+  tags?: Tag[]
+  price?: string
   event_start_at?: string
-}
-
-export interface ProjectDetail extends ListProject {
```
KaarelKa, line 209: **[blocker]**

Let's stick with ProjectDetail, the type was renamed to ProjectDetail on purpose, so it would be more consistent (We have SubProjectDetail, so ProjectDetail is more intuitive that DetailedProject). Basically we want naming to be intuitive and predictable, so every type that has a "Detail" and "List" version of it, we should keep the consistency. In this case it will be "ObjectNameDetail" instead of "DetailedObjectName". We can use the "DetailedObjectName" as well, but then it should be changed for every object

KaarelKa: Actually when we consider the naming structure for ListProject and ListSubProjectDetail, then better naming would be to have the object name in the end.

So ideally: 

1. ListSubProject
2. DetailedSubProject
3. ListProject
4. DetailedProject

But yeah not too important right now, we can deal with naming consistency a bit later

src/router/router.tsx
```diff
@@ -75,15 +75,12 @@ export const protectedRoutes: FullRouteObject[] = [
     path: 'projects',
     label: i18n.t('menu.projects'),
     Icon: ProjectsIcon,
-    privileges: [Privileges.ViewPersonalProject, Privileges.ViewPersonalTask],
+    privileges: [Privileges.ViewPersonalProject],
```
KaarelKa: **[blocker]**

We should remove the privileges part from this parent "projects" path completely, otherwise users without the ViewPersonalProject privilege can't see their tasks (which will be essentially all vendors + some PMs).

KaarelKa (changes requested):

PR #181, opened 2023-11-28
554

VaabelMaret:


VaabelMaret:
Note: One edge case remained unresolved, in the project adding tms modal.
There is a table with checkboxes. If you check the checkbox with keyboard the focus will go away from the modal. But if this checkbox is in the same table dropdown, the focus remains in this dropdown after checking.

KaarelKa:
Issue with search inside dropdown, I't seems I can't focus on it with just the keyboard

KaarelKa (approved):

PR #185, opened 2023-12-01
Improve-618-refetch project subprojects if project data changes

VaabelMaret:


src/hooks/requests/useProjects.ts
```diff
@@ -130,6 +130,9 @@ export const useUpdateProject = ({ id }: { id?: string }) => {
           return { data: newData }
         }
       )
+      map(data.sub_projects, (subProject) =>
```
KaarelKa, line 134: **[blocker]**

We should avoid refetching whenever possible. This should only be for cases, where we defiitely need it and have no other way. This one can cause a lot of refetching (1 API request per sub_project).

Inside the response we already have the updated subProjects, so we could just use that to update our cache, instead of refetching.

Extra note* current solution doesn't seem to work, since the new deadline_at value from the subproject is not added to the form. It will still show the old deadline for the subproject.

KaarelKa (changes requested):

src/hooks/requests/useProjects.ts
```diff
@@ -131,6 +131,18 @@ export const useUpdateProject = ({ id }: { id?: string }) => {
...
+        queryClient.setQueryData(
+          ['subprojects', subProject.id],
+          (oldData?: ProjectsResponse) => {
```
KaarelKa: **[blocker]**

Not sure about the typing here. oldData type should be SubProjectResponse

KaarelKa (changes requested):

PR #191, opened 2023-12-05
Improve - Improve - Move code for Assignment form to separate compone…

KaarelKa:
…nt, remove unused props, reduce unnecessary prop passing, fix for delete assignment button condition, fix console warnings


KaarelKa:
Also contains 378 task

PR #203, opened 2023-12-08
New-206-added vendor tasks page

VaabelMaret:


src/hooks/requests/useTasks.ts
```diff
@@ -21,11 +22,17 @@ export const useFetchTasks = (initialFilters?: TasksPayloadType) => {
   } = useFilters<TasksPayloadType>(initialFilters)
 
   const { isLoading, isError, data, refetch } = useQuery<TasksResponse>({
-    queryKey: ['tasks', filters],
+    queryKey: ['tasks'],
```
KaarelKa: **[blocker]**

Won't work exactly like this, since we have the "Minu ülesanded" and "Ootel ülesanded" lists in the same view. Currently it will show the same data for both.

src/pages/VendorTasks/VendorTasks.tsx
```diff
@@ -0,0 +1,48 @@
+import { FC } from 'react'
```
KaarelKa, line 1: **[blocker]**

We shouldn't allow navigating to those tasks for other users, only the vendor who is the assignee of the task should be able to view the task.

KaarelKa (changes requested):

PR #205, opened 2023-12-08
Improve - 505 - remove initial filter from add/remove vendors modal

KaarelKa:
Wrong branch name, it's actually 505 task


KaarelKa:
Added 632 task changes as well now

PR #219, opened 2023-12-15
NEW - 581 - Remember view filters while going back with breadcrumb or refreshing

rebekkabb:
Ticket - https://github.com/keeleinstituut/tv-tolkevarav/issues/581

To-do 
- This should be done by storing the selected filters on the URL and populating the table with the values from the URL on first load
- ATM we skipped all the modals
- Also moved the search bar inside of the tables in multiple spots

How to test - 
I guess you need to check all tables that we have and see if all the sorting, filtering, paginating is populated correctly on reload or breadcrumb redirection (and pressing back button as well)

And do let me know if you see way to simplify anything or if anything is misbehaving!


src/components/organisms/SideBar/SideBar.tsx
```diff
@@ -44,7 +44,11 @@ const MenuItems: FC<MenuItemsProps> = ({
       {map(
         menuItems,
         ({ children, path, Icon, isInterTitle, label, privileges }) => {
-          const fullPath = parentPath ? `${parentPath}/${path}` : `/${path}`
+          const fullPath = parentPath
+            ? path
```
KaarelKa, line 48: **[non-blocker]**

Would suggest avoiding ternary operators, if they have to be nested. The readability can get bad really quickly.

KaarelKa (approved):

PR #263, opened 2024-02-20
FIX-704 - Oral translation time and date issues

rebekkabb:
Ticket - https://github.com/keeleinstituut/tv-tolkevarav/issues/704

To-do - Oral translation time and date issues

Could not reproduce any of the error situations mentioned in the ticket

### DIFF
--- a/src/components/organisms/forms/AssignmentForm/AssignmentForm.tsx
+++ b/src/components/organisms/forms/AssignmentForm/AssignmentForm.tsx
@@ -81,17 +81,9 @@ const AssignmentForm: FC<AssignmentFormProps> = ({
     () => ({
       ...(deadline_at
         ? {
-            deadline_at: getLocalDateObjectFromUtcDateString(
-              (subProjectDeadline || '') < deadline_at
-                ? subProjectDeadline || ''
-                : deadline_at
-            ),
+            deadline_at: getLocalDateObjectFromUtcDateString(deadline_at),
           }
-        : {
-            deadline_at: getLocalDateObjectFromUtcDateString(
-              subProjectDeadline || ''
-            ),
-          }),
+        : {}),
       ...(shouldShowStartTimeFields && event_start_at
         ? {
             event_start_at: getLocalDateObjectFromUtcDateString(event_start_at),
@@ -108,7 +100,6 @@ const AssignmentForm: FC<AssignmentFormProps> = ({
       shouldShowStartTimeFields,
       volumes,
       assignee_comments,
-      subProjectDeadline,
     ]
   )
 
--- a/src/components/organisms/forms/AssignmentForm/AssignmentForm.tsx
+++ b/src/components/organisms/forms/AssignmentForm/AssignmentForm.tsx
@@ -1,4 +1,4 @@
-import { FC, useCallback, useMemo } from 'react'
+import { FC, useCallback, useEffect, useMemo } from 'react'
 import { useForm } from 'react-hook-form'
 import DynamicForm, {
   FieldProps,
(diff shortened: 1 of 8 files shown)


src/hooks/requests/useProjects.ts
```diff
@@ -193,6 +194,7 @@ export const useUpdateSubProject = ({ id }: { id?: string }) => {
           return { data: newData }
         }
       )
+      queryClient.refetchQueries({ queryKey: ['projects'] })
```
KaarelKa: **[blocker]**

Do we really need this ? we should avoid doing any unnecessary requests to BE. In most cases BE should return all the changed fields in the response of the PUT/POST request and we should be able to take the data from there and store it in our cache.

If something is missing, then first thing we should do is ask BE to add it and only if that is too much hassle for them, we can go with refetching.

**NB** Same comment applies for the refetch of subprojects after updating the project

src/components/organisms/forms/AssignmentForm/AssignmentForm.tsx
```diff
@@ -80,11 +80,21 @@ const AssignmentForm: FC<AssignmentFormProps> = ({
   const defaultValues = useMemo(
     () => ({
       ...(deadline_at
-        ? { deadline_at: getLocalDateOjectFromUtcDateString(deadline_at) }
-        : {}),
+        ? {
+            deadline_at: getLocalDateObjectFromUtcDateString(
```
KaarelKa: **[blocker]** (Possibly)

Are we sure about this ? Ideally the correct deadline_at value should be set in BE and we should display what is set there. Otherwise we can end up with a situation, where we show 1 deadline, but the actual deadline is different.

I know that the business logic states that assignment deadline can't be after subproject deadline, but this is something that BE should handle, so we wouldn't end up with faulty data.

**NB** If this has already been discussed with BE and the decision was to show it like this on FE, then feel free to ignore this comment

KaarelKa (changes requested):

src/hooks/requests/useProjects.ts
rebekkabb: First case: The date in the subproject tab is taken from the project which does not refresh when we update the date in subproject üldinfo, so we should ask for project info to be sent as well so we could refresh the project here?

Second case: The tasks get their deadline_at from the assigment cache, but the project subprojects here don't seem to include assigment ids so we should ask for ids and refresh the assignments that way?

KaarelKa: 1. The date you need is from the "project.deadline_at" or the deadline_at of 1 of the subprojects of that project ?

1.1. If it's the first one, then you should be able to get the changed value from the response of this update request. It has the "project" field with "deadline_at". Although I'm wondering what the case is, where any changes to the subProject affects the project.

1.2. If it's the other case, then you can have something like const deadlineToUse = innerDeadlineAt || deadline_at, where deadline_at is the one passed from props and innerDeadlineAt would be the deadline_at taken from subProject inside SubProjectSection. This way we will have the outer "deadline_at" for initial rendering and it will be replaced with innerDeadlineAt for all future cases (including, the case, where we update it)

2. Not exactly. Basically there are 2 options here.:

2.1. Ask if BE can return the changed assignments as well (in project.sub_projects)

2.2. Another option is to change them optimisitically. Essentially, if project deadline_at will change to sth greater than assignment deadline, then nothing changing for the assignments. If it changes into sth smaller, then the assignment deadline should change to project deadline. So essentially you could update the cache based on this.

rebekkabb: Thanks! Implemented 1.2 and 2.2 approaches, helped to nicely to refresh the view without having to refetch

src/components/organisms/forms/AssignmentForm/AssignmentForm.tsx
rebekkabb: Removed the logic from here and asked the back-end to implement the default (subproject) deadline for new tasks that we create by splitting. That part of the task we can test later when they have added the date.

KaarelKa (approved):

PR #267, opened 2024-02-27
FIX-712 - Fill review task assignee with manager info

rebekkabb:
Ticket - https://github.com/keeleinstituut/tv-tolkevarav/issues/712

To-do: Fill review task assignee with manager info (also in client view)

Currently approached with the understanding that we have just one PM that is assigned to the task (this goes for the client view and I also left out the removal button for the PM candidate)

### DIFF
--- a/src/components/molecules/AssignmentCandidatesSection/AssignmentCandidatesSection.tsx
+++ b/src/components/molecules/AssignmentCandidatesSection/AssignmentCandidatesSection.tsx
@@ -60,7 +60,8 @@ const AssignmentCandidatesSection: FC<AssignmentCandidatesSectionProps> = ({
           delete_button: vendor?.id,
         }
       })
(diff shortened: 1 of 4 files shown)


src/components/molecules/AssignmentCandidatesSection/AssignmentCandidatesSection.tsx
```diff
@@ -59,9 +59,21 @@ const AssignmentCandidatesSection: FC<AssignmentCandidatesSectionProps> = ({
           price,
           delete_button: vendor?.id,
         }
-      }),
-    [candidates]
-  )
+      })
+    } else if (manager_candidates?.length) {
```
KaarelKa: **[non-blocker]**

No need for "else if" and "else", when using "return" inside a function.

KaarelKa (approved):

PR #272, opened 2024-03-12
authserver

thenouan:


src/components/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,236 @@
...
+}
+
+const authContext = createContext<AuthContextType>(authContextDefaultValues)
+
+export const useAuth = () => {
+  return useContext(authContext)
+}
+
+export const rawLogin = () => {
+  window.location.href = `http://localhost:8000/login?redirect_uri=${window.location.href}`
```
KaarelKa: Hmm should be REACT_APP_GATEWAY_BASE from env ?

KaarelKa:
Can you also remove authApiClient for cleanup purposes. It's no longer used

src/components/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,236 @@
...
+}
+
+export const rawLogout = (error?: string) => {
+  let redirectUri = window.location.origin
+
+  if (error) {
+    redirectUri += `#${error}`
+  }
+  window.location.href = `http://localhost:8000/logout?redirect_uri=${encodeURIComponent(
```
KaarelKa: Same here

src/components/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,236 @@
...
+    redirectUri
+  )}`
+}
+
+export const AuthProvider: FC<PropsWithChildren> = (props) => {
+  const { children } = props
+
+  const queryClient = useQueryClient()
+
+  const login = useCallback(() => {
```
KaarelKa, line 92: I don't think you need to wrap rawLogin inside useCallback

thenouan: True, but I wanted to provide similar API for other components using values from auth context. Since both methods are quite similar, then I wanted them to go through same layers: global definition -> context definition -> usage in context consumers

src/components/contexts/AuthContext.tsx
thenouan: My missight, will change

src/components/contexts/AuthContext.tsx
thenouan: My missight, will change

src/components/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,265 @@
...
+    const now = Math.ceil(Date.now() / 1000)
+    const sessionExpires = Number(Cookies.get('session-expires'))
+
+    if (!sessionExpires) {
+      return
+    }
+    const remaining = sessionExpires - now
+    const isSessionExpired = now >= sessionExpires
+
+    const notificationThresholds = [3600, 1800, 60, 30].filter(
```
KaarelKa, line 200: Hmm, might be wrong, but I'm thinking that it would be simpler to just check if there is a notificationThreshhold that matches the current remaining

So basically replace this notificationThresholds, currentNotificationThreshold and if (remaining <= currentNotificationThreshold)

with

if (includes([3600, 1800, 60, 30], remaining))


Those seem to be the only cases, where we display the notification.
From the list of matched notificationThresholds, the only one which can possibly be >= remaining, is the one that is == remaining, since the others get filtered out.

Or I'm missing smth ?

KaarelKa: **Note***

Not really important, just a possibly way to simplify this a bit and reduce code

PR #281, opened 2024-04-25
New-717 Fix VendorsTable change when VendorsEditTable changes

IngaSokman:


KaarelKa:
Hmm currently visually it seems fine, but when looking at the network thab I see that we are actually doing 3 requests to BE every time the page is changed. 

1. The GET /institution-users with the correct params - This is correct
2. The GET /vendors with the page param from the institution users fetch (the modal) - This should not happen at all
3. The GET /vendors with the previous page param (correct page param) - This one makes sure that we still show the correct data.


Out of these the 2. and 3. should not happen at all. We are essentially doing 1 request that shouldn't happen and to fix the issue with data we do another request on top of it.


https://github.com/keeleinstituut/tv-web/assets/26221722/8cbed4c5-ae74-4a03-a1d4-b797cd8c5160

KaarelKa:
The original issue actually comes from the fact that handlePaginationChange doesn't have the saveParams check that is used to decide whether we should store the params in url or not. When dealing with modals we should not and the param is correctly passed in VendorsEditModal:

 useFetchUsers({
      initialFilters: {
        per_page: 10,
        page: 1,
      },
      saveQueryParams: false,
      useTranslationService: true,
    })
    
    saveQueryParams: false is correct, it just seems that it hasn't been fully implemented. handlePaginationChange should also check it

KaarelKa (changes requested):
*

rebekkabb:
The source of the issue is in TablePagination when navigating between pages we update the `page` and `per_page` query params from there. I added a flag for tables in modals so that we would know to skip this logic for them. Just in case added this flag to other modal tables as well.

KaarelKa (changes requested):
The solution is good, but let's remove a bit of prop drilling here. We have Modal context around all modals (inside MobalBase component). Every child has access to that context, so in reality every component knows whether it is inside a modal or not. We have many other components that also have differences based on whether they are inside a modal or not.

So let's also use the value from modal context for this logic. This will guarantee that every table inside a modal will handle this correctly and in the future we will not need to think about it, when adding new modals with new tables (there is always the risk that the developer dealing with it will forget).

For this you can use the useModalContext hook and check if the modalContentId exists or not from that hook. There are a bunch of example in the code for this already

PR #283, opened 2024-04-25
193

IngaSokman:


src/components/molecules/WorkingTimes/WorkingTimes.tsx
```diff
@@ -143,7 +170,10 @@ const WorkingTimes: FC<WorkingTimesPropType> = ({ data, id, name }) => {
         className={classes.editButton}
         icon={EditIcon}
         onClick={handleEditList}
-        hidden={!includes(userPrivileges, Privileges.EditInstitutionWorktime)}
+        hidden={
+          !includes(userPrivileges, Privileges.EditInstitutionWorktime) ||
```
KaarelKa: Hmm this privilege check might not work. Basically we are using this component in 2 places (Should be 3 I think). 

1. User page (So the page of any user)
2. Institution settings (For editing istitution working times)
3. User details (Currently seems to be commented out there, but we probably need to add it back, so the user can edit their own working times as well now that we have the endpoints for it).

With the current privilege check someone who has the privilege to edit user worktimes, but not the institution work times will still see this section under institution settings. Same is true for the opposite. User who can edit institution work times, but can't edit user work times will still see this section under User page.

Not sure what the permissions should be for editing your own worktime - This should be checked from Wiki.

KaarelKa (changes requested):

src/components/molecules/WorkingTimes/WorkingTimes.tsx
rebekkabb: I added a fix for the existing two cases, currently waiting for more information and clarity on the user details view permissions

rebekkabb: Added the working times to user details view. About privileges: basically we check if the user is editing themself, if not then we check for privilege

KaarelKa (approved):

PR #289, opened 2024-05-17
215 - Managing institution vacation times

rebekkabb:
~~For some reason form validation is not triggering on the datepicker inputs, I will keep investigating as I work on task 402, rest is ready for a look.~~ FIXED


src/components/molecules/DatePickerInput/DatePickerInput.tsx
```diff
@@ -58,10 +58,23 @@ const DatePickerComponent = ({
 
   const convertedValue = dayjs(value, 'DD/MM/YYYY')
   const splittedDayValue = convertedValue?.format('YYYY-MM-DD')
+  const calendarRef = useRef<any>(null)
```

We should have 0 uses of any type in the project. There might be some very specific edge cases, where it takes too much time to type sth and we fall back to any, but this shouldn't be 1 of them.

src/components/molecules/DatePickerInput/classes.module.scss
```diff
@@ -56,10 +56,6 @@
 .datePickerWrapper {
   width: 125px;
   position: relative;
-
-  &.increasedZIndex {
```
KaarelKa, line 60: **[blocker]**

I see that you removed this class from styles, but it's still used in html. Is there some workaround done, where increacing the z-index is not necessary anymore inside modals ? If yes, then we should also remove all uses of this class name.

src/components/molecules/VacationTimes/VacationTimes.tsx
```diff
@@ -0,0 +1,107 @@
...
+import dayjs from 'dayjs'
+import timezone from 'dayjs/plugin/timezone'
+import { useAuth } from 'components/contexts/AuthContext'
+import { Privileges } from 'types/privileges'
+import { EditDataType } from 'components/organisms/modals/DateRangeFormModal/DateRangeFormModal'
+
+dayjs.extend(timezone)
+interface VacationTimesPropType {
+  data?: InstitutionVacationType[]
+  name: string
```
KaarelKa: **[blocker]**

It seems that name and id are typed and passed to this component, but not used

src/hooks/requests/useInstitutions.ts
```diff
@@ -106,3 +107,43 @@ export const useUpdateInstitutionDiscounts = () => {
...
+    isLoading: isLoading,
+    isError: isError,
+  }
+}
+
+export const useInstitutionVacationsUpdate = () => {
+  const queryClient = useQueryClient()
+  const { mutateAsync: updateInstitutionVacations, isLoading } = useMutation({
+    mutationKey: ['institution-vacations'],
+    mutationFn: (payload: any) =>
```
KaarelKa: **[blocker]**

"any" type should be changed to actual type

src/hooks/requests/useInstitutions.ts
```diff
@@ -106,3 +107,43 @@ export const useUpdateInstitutionDiscounts = () => {
...
+  const { isLoading, isError, data } = useQuery<InstitutionVacationsDataType>({
+    queryKey: ['institution-vacations'],
+    queryFn: () => apiClient.get(`${endpoints.INSTITUTION_VACATIONS}`),
+  })
+
+  const { data: institutionVacations } = data || {}
+
+  return {
+    institutionVacations,
+    isLoading: isLoading,
```
KaarelKa: **[non-blocker]**

You can use shorthand here as well. If you have a variable with the same name as the key name you want for an object, then you can just write 

 return {
    institutionVacations,
    isLoading,
    isError,
  }

src/pages/InstitutionSettings/InstitutionSettings.tsx
```diff
@@ -38,8 +43,13 @@ const InstitutionSettings: FC = () => {
           workingTimes={
             <WorkingTimes name={name} id={institutionId} data={institution} />
           }
-          //TODO: add vacation days component as prop
-          //vacationDays={<div>Puhkepäevad:</div>}
+          vacationDays={
```
KaarelKa: **[non-blocker]**

Not really related to this task, but wondering why workingTimes and vacationDays are added as props to InstitutionForm, instead of just importing them there. InstitutionForm is not used anywhere else and never has any other components for these props

KaarelKa (changes requested):

src/components/molecules/DatePickerInput/classes.module.scss
rebekkabb: Indeed I forgot to remove the use of this, I removed this class because in the modal all the inputs were on top of the opened date picker calendars. Overall everything looks good in modals without this class

src/pages/InstitutionSettings/InstitutionSettings.tsx
rebekkabb: Yep interesting, maybe there were some different plans earlier, regardless we are importing them from InstitutionForm now

src/components/organisms/forms/InstitutionForm/InstitutionForm.tsx
```diff
@@ -174,8 +173,8 @@ const InstitutionForm: FC<InstitutionPropTypes> = ({
         />
 
         <div className={classes.dateTimeContainer}>
-          {workingTimes}
-          {vacationDays}
+          <WorkingTimes name={name} id={id} data={props} />
```
KaarelKa, line 172: **[non-blocker]**


Currently name and id are passed twice (They are also passed inside data={props}). Another part is that the "props" part is 1 level deeper than it needs to be. Generally it's good to keep props as shallow as possible.

In this case I would pass {...props} to WorkingTimes and inside the WorkingTimes component they could be destructured in the following way:

{ id, name, ...data } = props

src/pages/InstitutionSettings/InstitutionSettings.tsx
```diff
@@ -20,6 +22,7 @@ const InstitutionSettings: FC = () => {
   const { institution } = useInstitutionFetch({
     id: institutionId,
   })
+  const { institutionVacations } = useInstitutionVacationsFetch()
```
KaarelKa: **[non-blocker]**

This could also be inside InstitutionForm. Seems that institutionVacations is only needed inside InstitutionForm

KaarelKa (approved):

PR #333, opened 2026-01-09
918 update code deps

kadmit:
Task: https://github.com/keeleinstituut/tv-tolkevarav/issues/918

I manage to solve all vulnarability issues but the code still has outdated dependencies:

Screenshot 2026-01-09 at 14 48 00


kadmit:
I manage to decrease amount of outdated dependencies:

Screenshot 2026-01-27 at 10 09 20

src/reportWebVitals.ts
kadmit, line 1: The change comes from updating of  `web-vitals` to `^5.1.0`.

eslint.config.cjs
kadmit, line 1: The eslint configuration is changed as previously the application used "react-app" plugin (comes from CRA) which is outdated. 

Similar topic: https://github.com/facebook/create-react-app/issues/13525

